PR #341, opened 2026-04-24
test(orchestration): critical coverage for pipeline_utils, workflows, workflow_dsl (#329 Track 2)

jsboigeEpita:
## Summary
- Adds 39 tests covering hot-path callables in `argumentation_analysis/orchestration/`
- Targets uncovered branches in `pipeline_utils.py`, `workflows.py`, and `workflow_dsl.py`

## Coverage Impact
| Module | Before | After |
|--------|--------|-------|
| `pipeline_utils.py` | 84% | 84% (edge case robustness) |
| `workflow_dsl.py` | 74% | **81%** (+7%) |
| `workflows.py` | 67% | **76%** (+9%) |

## Test Categories
- **pipeline_utils**: TTL never-expire, cache clear/cleanup, FIFO eviction, `_MetricsContext` manager, batch `fail_fast`
- **workflows**: `_should_rerun_fallacy` (7 edge cases), `_increment_fallacy_rerun`, all 7 workflow builders, catalog lifecycle
- **workflow_dsl**: condition eval (true/false/error), input_transform (apply/fallback), loop (max_iter/early convergence), state writer (success/error)

## Test plan
- [x] All 39 tests pass locally (`pytest tests/unit/argumentation_analysis/orchestration/test_critical_coverage.py -v`)
- [x] Coverage measured with `--cov-report=term-missing`
- [x] No regressions in existing orchestration tests

Ref: #329 Track 2

### DIFF
--- a/tests/unit/argumentation_analysis/orchestration/test_critical_coverage.py
+++ b/tests/unit/argumentation_analysis/orchestration/test_critical_coverage.py
@@ -1,0 +1,565 @@
+"""Coverage improvements for critical orchestration and evaluation modules.
+
+#329 Track 2: Adds tests for uncovered hot-path callables in:
+- pipeline_utils.py (edge cases: never-expire, clear, cleanup, metrics context)
+- workflows.py (_should_rerun_fallacy, iterative workflow, catalog, reset)
+- workflow_dsl.py (condition, input_transform, loop, state writer)
+"""
+
+import asyncio
+import time
+
+import pytest
+
+from argumentation_analysis.orchestration.pipeline_utils import (
+    AnalysisCache,
+    CacheEntry,
+    PipelineMetrics,
+    _MetricsContext,
+    BatchRequest,
+    BatchResult,
+    run_batch_analysis,
+)
+from argumentation_analysis.orchestration.workflows import (
+    _should_rerun_fallacy,
+    _increment_fallacy_rerun,
+    build_iterative_analysis_workflow,
+    build_nl_to_logic_workflow,
+    build_quality_gated_counter_workflow,
+    build_debate_governance_loop_workflow,
+    build_jtms_dung_loop_workflow,
+    build_neural_symbolic_fallacy_workflow,
+    build_hierarchical_fallacy_workflow,
+    get_workflow_catalog,
+    reset_workflow_catalog,
+)
+from argumentation_analysis.orchestration.workflow_dsl import (
+    WorkflowBuilder,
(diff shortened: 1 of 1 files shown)


jsboigeEpita (commented):
## PR #341 Review — Critical Coverage for Pipeline Utils, Workflows, Workflow DSL (#329 Track 2)

**Verdict: APPROVE**

565 lines of well-targeted tests covering uncovered hot-path callables across 3 modules. Each test class maps to a specific module with clear coverage targets.

### What's good

1. **`TestCacheEntryNeverExpires`** — Tests the edge case where `ttl_seconds=0` and negative TTLs should never expire. Correct boundary testing.

2. **`TestAnalysisCacheCleanup`** — Uses `time.sleep(0.08)` with a 0.05s TTL. This is a real timing test that could be flaky in slow CI. The tolerance (0.08s vs 0.05s) is generous enough to work in practice but tight enough to be meaningful.

3. **`TestMetricsContextManager`** — Tests success, failure (via `pytest.raises`), metadata passthrough, and missing confidence. The empty summary test is a good null-case.

4. **`TestShouldRerunFallacy`** — 7 edge cases including: no retractions, positive undermined count, invalid beliefs, max count boundary, non-dict jtms_output, non-dict beliefs. Thorough.

5. **`TestWorkflowCondition`** — Tests false (skip), true (run), and exception in condition (should proceed — verifies error handling).

6. **`TestWorkflowLoop`** — Tests both max-iteration exhaustion and early convergence. The convergence function `curr.get("iteration", 0) >= 2` is clear.

7. **`TestWorkflowStateWriter`** — Tests successful write and write failure (should not fail the phase).

### Notes (non-blocking)

- **Import in test body** (line 279): `from argumentation_analysis.orchestration.workflows import WORKFLOW_CATALOG` is imported inside `test_reset_clears_catalog`. This works but is unusual — if the module-level import already imports the catalog, verify this doesn't create a different binding than the one used by `get_workflow_catalog()`. If `WORKFLOW_CATALOG` is a module-level dict, both references should point to the same object.

- **`mock_state` type creation** (lines 522, 557): `type("State", (), {...})()` creates a mock class inline. This works but could be replaced with `MagicMock(spec_set=["log_error", "set_workflow_results"])` for clarity. Minor style point.

- **Batch fail-fast test** (line 165): The assertion `assert len(results) == 2` verifies that even with fail-fast, both results are returned (one as failure). Good — confirms fail-fast doesn't silently drop results.

PR #343, opened 2026-04-24
fix(tests): increase timeout in flaky test_timeout_raises_error

jsboigeEpita:
## Summary
- Increase `timeout` from 0.2s to 0.5s in `test_timeout_raises_error`

## Context
The test uses `threading.Event.wait(timeout=...)` to verify that a request timeout raises `RequestTimeoutError`. The 0.2s value was too tight and caused intermittent failures during full suite runs (2845 tests) on slower machines. 0.5s provides sufficient headroom while keeping the test fast.

## Test plan
- [x] Test passes in isolation
- [x] Test passes under full suite load
- [x] No behavior change to production code

### DIFF
--- a/tests/unit/argumentation_analysis/core/communication/test_request_response.py
+++ b/tests/unit/argumentation_analysis/core/communication/test_request_response.py
@@ -352,7 +352,7 @@ class TestSendRequest:
                 recipient="b",
                 request_type="q",
                 content={},
-                timeout=0.2,
+                timeout=0.5,
                 retry_count=0,
             )
 


jsboigeEpita (commented):
## PR #342, #343 — Timeout Increases for Benchmark Scripts and Flaky Test

Reviewing together as they address the same problem: timeouts too tight for CI.

### #342 — Benchmark timeout defaults 120s/180s → 600s (4+, 4-, 2 files)

Changes default timeout in two benchmark scripts:
- `run_benchmark.py`: `120.0` → `600.0` (both function default and argparse default)
- `run_benchmark_multimodel.py`: `180.0` → `600.0` (same)

**Verdict: APPROVE** — The 5x increase is reasonable for CI environments where LLM calls can be slow. Users can still override via `--timeout` flag.

### #343 — Flaky test timeout 0.2s → 0.5s (1+, 1-, 1 file)

Changes `timeout=0.2` to `timeout=0.5` in `test_timeout_raises_error` in `test_request_response.py`.

**Verdict: APPROVE** — A 0.2s timeout is indeed flaky in CI. 0.5s gives enough margin while still testing that timeouts actually fire. The test name confirms it's specifically testing timeout behavior, so the timeout value itself isn't what's being asserted.

Issue #347, opened 2026-04-25
[1.1] Spectacular workflow profile


jsboigeEpita:
**Parent Epic:** #344
**Track:** 1 — Spectacular Analysis Pipeline (scaffolding)

## Goal
Créer un nouveau profile de workflow `spectacular` qui compose toute la chaîne : extraction → formal logic (PL/FOL/Modal) → 3-tier fallacy → JTMS → ATMS → Dung → ASPIC → counter-arg → debate → governance → quality → synthesis narrative.

## Files to touch
- `argumentation_analysis/orchestration/workflows.py` — ajouter `build_spectacular_workflow()`
- `argumentation_analysis/orchestration/registry_setup.py` — register workflow
- `argumentation_analysis/run_orchestration.py` — exposer `--workflow spectacular`

## Definition of Done
- [ ] `--workflow spectacular` runs end-to-end sur doc_A sans erreur
- [ ] State JSON contient ≥ 28/32 champs
- [ ] Tous les phases du DAG s'exécutent (pas de skip)
- [ ] Test unitaire `test_spectacular_workflow_dag.py` valide la structure
- [ ] PR mergée, CI green

## Dependencies
**Blocks:** issues #1.2, #1.3, #1.4, #1.5, #5.1, #5.2, #5.3.
**Depends on:** #0.A (audit informe le wiring).

## Assignee
po-2025

PR #376, opened 2026-04-30
feat(atms): multi-context hypothesis activation (#349)

### LINKED ISSUE #349: [1.3] ATMS multi-context activation
**Parent Epic:** #344
**Track:** 1 — Spectacular Analysis Pipeline

## Goal
Activer ATMS dans le workflow spectacular avec 3-4 hypothèses testées et cohérence par contexte rendue.

## Background
ATMS existe (`atms_core.py`, `atms_plugin.py` + tests + factory/registry) mais pas trop approfondi. À exploiter pour montrer raisonnement multi-mondes.

## Files to touch
- `argumentation_analysis/services/jtms/atms_core.py` — vérifier API
- `argumentation_analysis/plugins/atms_plugin.py` — wiring
- `argumentation_analysis/orchestration/invoke_callables.py` — `_invoke_atms`
- `argumentation_analysis/orchestration/state_writers.py` — `atms_contexts` field

## Definition of Done
- [ ] doc_A : state.atms_contexts contient 3-4 hypothèses
- [ ] Pour chaque contexte, cohérence calculée (coherent / incoherent)
- [ ] Au moins 1 hypothèse cohérente sous certains contextes, incohérente sous d'autres
- [ ] Test `test_atms_multi_context.py` passe
- [ ] PR mergée, CI green

## Dependencies
**Depends on:** #1.1
**Parallel with:** #1.2, #1.4, #1.5

## Assignee
po-2025

jsboigeEpita:
## Summary
- **_generate_hypotheses()**: Builds 3-4 varied assumption sets (full trust, fallacy-excluded, high-quality, skeptical) from upstream analysis data
- **_invoke_atms multi-context**: Tests each hypothesis against the ATMS for consistency, computing derivable/contradicting beliefs per context
- **state.atms_contexts**: New `UnifiedAnalysisState` field storing hypothesis coherence results with per-context `coherent` boolean
- **_write_atms_to_state**: Persists `atms_contexts` from ATMS output to state

## DoD Checklist (#349)
- [x] `state.atms_contexts` contains 3-4 hypotheses
- [x] Each context has coherence computed (`coherent` / incoherent)
- [x] At least 1 hypothesis coherent under some contexts, incoherent under others
- [x] `test_atms_multi_context.py` passes (14 tests)
- [ ] CI green

## Test plan
- [x] `pytest tests/unit/argumentation_analysis/test_atms_multi_context.py -v` — 14/14 passed
- [ ] Existing ATMS tests unaffected

Closes #349

### DIFF
--- a/argumentation_analysis/core/shared_state.py
+++ b/argumentation_analysis/core/shared_state.py
@@ -411,6 +411,8 @@ class UnifiedAnalysisState(RhetoricalAnalysisState):
         self.formal_synthesis_reports: List[Dict[str, Any]] = []
         # NL-to-formal-logic translations (#173)
         self.nl_to_logic_translations: List[Dict[str, Any]] = []
+        # ATMS multi-context analysis (#349)
+        self.atms_contexts: List[Dict[str, Any]] = []
         # Workflow execution results
         self.workflow_results: Dict[str, Any] = {}
 
--- a/argumentation_analysis/orchestration/invoke_callables.py
+++ b/argumentation_analysis/orchestration/invoke_callables.py
@@ -33,6 +33,7 @@ __all__ = [
     "_invoke_governance",
     "_invoke_jtms",
     "_invoke_atms",
+    "_generate_hypotheses",
     "_invoke_camembert_fallacy",
     "_invoke_local_llm",
     "_invoke_semantic_index",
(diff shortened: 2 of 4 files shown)


jsboigeEpita (commented):
# PR Review: #374, #375, #376 (Intelligence-Symbolique)

---

## PR #374: `feat(orchestration): spectacular workflow profile (#347)` -- 281 lines

**Verdict: APPROVE**

### Summary

Adds a new `build_spectacular_workflow()` that composes all available capabilities into a single full-chain DAG workflow. Registers it in the catalog under the `"spectacular"` key. Well-structured, follows existing patterns exactly.

### Positive observations

1. **DAG design is sound.** The dependency graph is a valid DAG with `extract` as the sole root. Parallelism is exploited correctly -- L1 phases (quality, nl_to_logic, neural_detect, hierarchical_fallacy) all depend only on extract and can run concurrently. The critical path goes through nl_to_logic -> pl -> dung_extensions -> aspic_analysis -> formal_synthesis (L0-L2-L3-L4-L8), while quality branches independently into counter -> jtms/debate -> atms.

2. **Optional phases are marked correctly.** Formal logic layers (L2), neural detection, hierarchical fallacy, and most downstream phases are `optional=True`. Only `extract`, `quality`, and `counter` are mandatory, which is appropriate -- the workflow degrades gracefully when optional capabilities aren't registered.

3. **Test coverage is thorough.** 14 tests covering: build succeeds, phase count (16), all phase names present, capability uniqueness, DAG validation, execution order resolvability, extract-is-root, specific dependency chains (formal logic -> nl_to_logic, dung -> fallacy+pl, atms -> jtms, formal_synthesis aggregation), catalog registration, comparison with standard workflow, capability presence checks.

4. **Catalog registration follows the exact existing pattern.** The `"spectacular"` key is added alongside existing keys in `get_workflow_catalog()` with no structural deviation.

5. **The `autouse=True` fixture for `reset_workflow_catalog()`** is correct and matches what other workflow tests do -- prevents catalog state leaking between tests.

### Minor observations (non-blocking)

- **Docstring mentions "target >=28/32" UnifiedAnalysisState field coverage.** This is an aspirational design goal, not something enforced in code. Fine as documentation, but a comment or assertion in a test verifying actual state field coverage would be more valuable.

- **No test for `get_required_capabilities()` including all expected capabilities.** The test `test_spectacular_includes_modal_logic` and `test_spectacular_includes_atms` check two specific capabilities but don't verify the full set. Low risk given the phase count test, but a comprehensive capability-set test would be a safety net.

---

## PR #375: `feat(logic): FOL signature pre-declaration (#348)` -- 239 lines

**Verdict: REQUEST CHANGES**

### Summary

Adds `extract_fol_metadata()` and `build_signature_prefixed_formulas()` static methods to `FOLLogicAgent` for parsing FOL formulas into Tweety-compatible sort/type declarations. Integrates into `_invoke_fol_reasoning` and the state writer. The goal is sound -- Tweety's FolParser needs pre-declared signatures. However, there are correctness issues in the regex parsing logic and dead code in the state writer.

### Issues

#### 1. Regex-based FOL parsing is fragile (correctness risk)

```python
for match in re.finditer(r"([A-Z][A-Za-z_]*)\(([^)]+)\)", formula):
```

This pattern matches `Predicate(arg1, arg2)` which is fine for simple formulas, but will produce incorrect results on several common FOL patterns:

- **Nested predicates:** `P(Q(x), y)` -- the `[^)]+` pattern will match `Q(x` as a single argument string, then split by comma produces `["Q(x", " y"]`. The arg `Q(x` starts with uppercase so it gets classified as a variable when it's actually a predicate application. The extracted arity of `P` will be wrong (2 instead of 2, but the arg classification is wrong).

- **Negation wrapping:** `!Mortel(socrate)` -- works, the regex finds `Mortel(socrate)` correctly.

- **Implication inside predicate args:** Less common but `R(x => y, z)` would break the comma split.

- **Predicate names starting with lowercase:** The regex requires `[A-Z]` as the first character. Standard FOL conventions vary. If the LLM generates `mortal(socrates)` this is silently missed. This may be intentional given the codebase convention of capitalizing predicates, but it should be documented.

**Recommendation:** At minimum, document the parsing assumptions in the docstring (e.g., "Expects predicates to start with uppercase, constants lowercase, no nested predicate arguments"). Consider adding a test case with a negated formula like `!Mortel(socrate)` and a universally quantified one like `forall X: (Homme(X) => Mortel(X))` to verify behavior under those patterns. The quantified test exists and passes, which is good.

#### 2. Empty-string guard missing in arg classification

```python
for arg in args:
    if arg[0].isupper():
        variables.add(arg)
    else:
        constants.add(arg)
```

If `args_str` contains trailing/leading whitespace after stripping, or if `args` ends up with an empty string (e.g., `P(, x)` -- malformed but LLM output), `arg[0]` raises `IndexError`. Add a guard:

```python
for arg in args:
    arg = arg.strip()
    if not arg:
        continue
    if arg[0].isupper():
        variables.add(arg)
    else:
        constants.add(arg)
```

#### 3. Dead code in state writer

```python
# state_writers.py
if not hasattr(state, "fol_signature"):
    state.fol_signature = fol_signature
else:
    state.fol_signature = fol_signature
```

Both branches are identical. This should just be `state.fol_signature = fol_signature`. The `hasattr` check is redundant since `fol_signature` is initialized in `__init__` (added in `shared_state.py` in this same PR).

#### 4. Type annotation for `constants` and `variables`

```python
constants: set = set()
variables: set = set()
```

These should be `Set[str]` for clarity and to match the `predicates: Dict[str, int]` annotation above.

#### 5. Duplicate signature extraction in fallback path

In the `except Exception` fallback of `_invoke_fol_reasoning`, the signature extraction is repeated:

```python
try:
    from argumentation_analysis.agents.core.logic.fol_logic_agent import FOLLogicAgent
    meta = FOLLogicAgent.extract_fol_metadata(formulas)
    fol_signature = meta.get("signature_lines", [])
except Exception:
    pass
```

This is the same import and call as the happy path. Consider extracting it to a local helper or at minimum using the already-imported `FOLLogicAgent` from the outer scope (it's imported at the top of the try block). Actually, the import is inside the try block and the fallback is in the except, so the name might not be bound. Still, the pattern is correct for safety -- just worth noting the duplication.

### Positive observations

- **Test quality is good.** Covers empty formulas, single predicate, quantified formulas, multi-arg predicates, signature line format, mixed formulas, cross-formula constant collection, and the `build_signature_prefixed_formulas` wrapper.
- **Shared state field is properly initialized** in `__init__`.
- **The `build_signature_prefixed_formulas` method** correctly returns original formulas unchanged when no metadata is extractable.

---

## PR #376: `feat(atms): multi-context hypothesis activation (#349)` -- 405 lines

**Verdict: REQUEST CHANGES**

### Summary

Extends `_invoke_atms` to generate 3-4 hypotheses from argument/fallacy data, test each against the ATMS for consistency, and report derivable/contradicting beliefs per context. Adds `_generate_hypotheses()` as a standalone helper and stores results in `state.atms_contexts`. The feature is valuable but has a subtle logic bug in the hypothesis matching and a weak differential coherence test.

### Issues

#### 1. Fuzzy argument name matching in hypothesis 3 (correctness bug)

```python
# Hypothesis 3: Only high-quality arguments
for arg_name in arg_names:
    for arg_id, scores in per_arg_scores.items():
        if isinstance(scores, dict) and arg_name[:30] in str(arg_id):
```

The `arg_name[:30]` prefix matching against `str(arg_id)` is dangerously loose. If `arg_name` is `"arg"`, then `"arg"[:30]` is `"arg"`, and the `in` check will match any `arg_id` containing the substring `"arg"` -- including `"counter_argument"`, `"target_argument"`, or `"argument_quality"`. This would cause arguments to be incorrectly classified as high-quality based on unrelated score entries.

**Recommendation:** Use exact matching or at minimum full-string containment with word boundary awareness:

```python
if isinstance(scores, dict) and arg_name in str(arg_id):
```

Or better, if `per_arg_scores` keys are meant to correspond to `arg_names`, use direct lookup:

```python
if isinstance(per_arg_scores.get(arg_name), dict):
    scores = per_arg_scores[arg_name]
    if scores.get("overall", scores.get("note_finale", 0)) >= 3.0:
        high_quality.append(arg_name)
```

#### 2. Truncation of fallacy target strings silently drops matches

```python
fallacy_targets.add(str(target)[:60])
```

Then later:

```python
clean_args = [a for a in arg_names if a not in fallacy_targets]
```

If `arg_names` contain entries longer than 60 characters, the truncated `target` will never match them, and fallacy-implicated arguments will incorrectly survive in the `clean_args` list. If this truncation is a safety measure, it should be documented. If it's intentional because `arg_names` are known to be short, then it's misleading -- better to truncate `arg_names` to the same length when comparing, or use a different matching strategy.

#### 3. Weak differential coherence test

```python
def test_differential_coherence(self):
    ...
    assert len(coherent) >= 1 or len(incoherent) >= 1
```

This assertion is a tautology -- with 3+ hypotheses, it is nearly impossible for all of them to be neither coherent nor incoherent (they all have a boolean `coherent` field). The test will pass even if every hypothesis is coherent, which would mean the multi-context feature provides no discriminating value. The docstring says "At least one hypothesis coherent, another incoherent when fallacies present" but the assertion only requires one or the other.

**Recommendation:** Strengthen to:

```python
assert len(coherent) >= 1 and len(incoherent) >= 1
```

Or at minimum, when `has_contradictions` is True, assert that `h_full_trust` is incoherent (similar to `test_context_with_fallacies_has_incoherent_hypothesis` but without the `if result["has_contradictions"]` guard that makes it a no-op when things go wrong).

#### 4. Deprecated `asyncio.get_event_loop()` usage

```python
def _run(coro):
```

`asyncio.get_event_loop()` is deprecated since Python 3.10 and will emit warnings in 3.12+. The modern replacement is:

```python
def _run(coro):
    return asyncio.run(coro)
```

This matches the broader ecosystem trend and is simpler.

#### 5. `sorted(set(...))` on derivable/contradicting beliefs is unnecessary

```python
"derivable_beliefs": sorted(set(derivable_beliefs)),
```

The `derivable_beliefs` list is built by appending unique node names (the `break` after each match prevents duplicates within a single node's environments, but the same belief could appear from different nodes). The `set()` dedup is correct, but the docstring or code should clarify whether duplicate beliefs across nodes are expected.

### Positive observations

- **`_generate_hypotheses` is cleanly separated** as a pure function with no side effects, making it independently testable.
- **Hypothesis design is thoughtful:** full trust, fallacy-excluded, quality-filtered, and skeptical contexts cover the meaningful dimensions of argumentation analysis.
- **The `has_contradictions` rewrite** from a specific node lookup to a filtered iteration is more robust:
  ```python
  # Old (fragile):
  len(atms.nodes.get("⊥", ATMSNode("", False)).label) > 0
  # New (clearer):
  any(len(n.label) > 0 for name, n in atms.nodes.items() if name == "⊥")
  ```
  The old version silently created a default `ATMSNode("", False)` on miss, which could mask bugs. The new version is explicit.
- **State writer correctly initializes** `atms_contexts` field in `__init__` and stores the data from output.
- **Guard for minimum 3 hypotheses** ensures the feature delivers meaningful multi-context analysis even with sparse input data.

PR #378, opened 2026-04-30
feat(jtms): retraction cascade rendering (#350)

### LINKED ISSUE #350: [1.4] JTMS retraction cascade rendering
**Parent Epic:** #344
**Track:** 1 — Spectacular Analysis Pipeline

## Goal
Surfacer dans le state les chaînes de rétractions cascadées causées par la détection de fallacies, dans un format lisible humainement.

## Background
Round 80 : JTMS retraction propagation fixée mais output ad-hoc. Cible : sequence "sophism detected → belief X retracted → cascaded retraction of Y because Y depended on X".

## Files to touch
- `argumentation_analysis/services/jtms/jtms_core.py` — expose retraction trace
- `argumentation_analysis/orchestration/state_writers.py` — `jtms_retraction_chain` field

## Definition of Done
- [ ] doc_A : state.jtms_retraction_chain contient au moins 1 chaîne de cascade complète
- [ ] Format : list de dicts `{trigger: ..., retracted: [...], cascaded: [...], reason: ...}`
- [ ] Test `test_jtms_cascade_render.py` passe avec scenario de cascade
- [ ] PR mergée, CI green

## Dependencies
**Depends on:** #1.1
**Parallel with:** #1.2, #1.3, #1.5

## Assignee
po-2025

jsboigeEpita:
## Summary
- **JTMS core tracing**: `enable_tracing()` / `get_retraction_chain()` capture cascading retractions when beliefs are invalidated
- **Cascade tracking**: `Belief.propagate()` records which downstream beliefs lose validity via `_jtms_ref`
- **Pipeline integration**: `_invoke_jtms` enables tracing before fallacy retraction, returns `retraction_chain`
- **State field**: `state.jtms_retraction_chain` stores human-readable cascade entries

## DoD Checklist (#350)
- [x] `state.jtms_retraction_chain` contains at least 1 complete cascade chain
- [x] Format: list of dicts `{trigger: ..., retracted: [...], cascaded: [...], reason: ...}`
- [x] `test_jtms_cascade_render.py` passes with cascade scenario (13 tests)
- [ ] CI green

## Test plan
- [x] `pytest tests/unit/argumentation_analysis/test_jtms_cascade_render.py -v` — 13/13 passed
- [ ] Existing JTMS tests unaffected

Closes #350

### DIFF
--- a/argumentation_analysis/core/shared_state.py
+++ b/argumentation_analysis/core/shared_state.py
@@ -385,6 +385,8 @@ class UnifiedAnalysisState(RhetoricalAnalysisState):
         self.argument_quality_scores: Dict[str, Dict[str, Any]] = {}
         # JTMS belief network (1.4.1)
         self.jtms_beliefs: Dict[str, Dict[str, Any]] = {}
+        # JTMS retraction cascade chains (#350)
+        self.jtms_retraction_chain: List[Dict[str, Any]] = []
         # Dung abstract argumentation frameworks (abs_arg_dung)
         self.dung_frameworks: Dict[str, Dict[str, Any]] = {}
         # Governance decisions and votes (2.1.6)
--- a/argumentation_analysis/orchestration/invoke_callables.py
+++ b/argumentation_analysis/orchestration/invoke_callables.py
@@ -989,6 +989,7 @@ async def _invoke_jtms(input_text: str, context: Dict[str, Any]) -> Dict:
 
     # ── Step 4: Fallacies → retract undermined beliefs + propagation ─
     fallacy_beliefs = []
+    session.jtms.enable_tracing()  # Track retraction cascades (#350)
     for i, f in enumerate(detected_fallacies[:6]):
         if not isinstance(f, dict):
             continue
@@ -1138,6 +1139,7 @@ async def _invoke_jtms(input_text: str, context: Dict[str, Any]) -> Dict:
         "formal_consistency": formal_consistency,
         "session_version": session.version,
         "consistency_checks": session.consistency_checks,
+        "retraction_chain": session.jtms.get_retraction_chain(),
     }
 
 
--- a/argumentation_analysis/orchestration/state_writers.py
+++ b/argumentation_analysis/orchestration/state_writers.py
@@ -171,6 +171,10 @@ def _write_jtms_to_state(output, state, ctx) -> None:
         if not isinstance(justifications, list):
             justifications = []
         state.add_jtms_belief(str(name), valid, justifications=justifications)
+    # Store retraction cascade chains (#350)
+    retraction_chain = output.get("retraction_chain", [])
+    if isinstance(retraction_chain, list):
+        state.jtms_retraction_chain = retraction_chain
 
 
 def _write_atms_to_state(output, state, ctx) -> None:
--- a/argumentation_analysis/services/jtms/jtms_core.py
+++ b/argumentation_analysis/services/jtms/jtms_core.py
@@ -93,7 +93,21 @@ class Belief:
         self._propagating = True
         try:
             for justification in self.implications:
-                justification.conclusion.compute_truth_statement()
+                conclusion = justification.conclusion
+                old_valid = conclusion.valid
+                conclusion.compute_truth_statement()
+                # Track cascade: conclusion was valid, now retracted
+                if (
+                    hasattr(self, '_jtms_ref')
+                    and self._jtms_ref is not None
+                    and self._jtms_ref._tracing_enabled
+                    and old_valid is True
+                    and conclusion.valid is not True
+                    and self._jtms_ref._retraction_trace
+                ):
+                    self._jtms_ref._retraction_trace[-1]["cascaded"].append(
+                        conclusion.name
+                    )
         finally:
             self._propagating = False
 
@@ -131,11 +145,15 @@ class JTMS:
     def __init__(self, strict: bool = False):
         self.beliefs: Dict[str, Belief] = {}
         self.strict = strict
+        self._retraction_trace: List[Dict] = []
+        self._tracing_enabled: bool = False
 
     def add_belief(self, name: str):
         """Add a new belief to the system."""
         if name not in self.beliefs:
-            self.beliefs[name] = Belief(name)
+            belief = Belief(name)
+            belief._jtms_ref = self
+            self.beliefs[name] = belief
 
     def remove_belief(self, belief_name: str):
         """Remove a belief and clean up its implications."""
@@ -153,7 +171,16 @@ class JTMS:
         """Set the truth value of a belief and propagate."""
         if belief_name not in self.beliefs:
             raise KeyError(f"Unknown belief: {belief_name}")
-        self.beliefs[belief_name].set_truth_value(validity)
+        belief = self.beliefs[belief_name]
+        old_valid = belief.valid
+        if self._tracing_enabled and old_valid is True and validity is not True:
+            self._retraction_trace.append({
+                "trigger": belief_name,
+                "retracted": [belief_name],
+                "cascaded": [],
+                "reason": f"directly set to {validity}",
+            })
+        belief.set_truth_value(validity)
 
     def add_justification(
         self,
@@ -206,6 +233,19 @@ class JTMS:
         """Print all beliefs and their truth values."""
         for b in self.beliefs.values():
             print(b)
+
+    def enable_tracing(self):
+        """Enable retraction cascade tracing."""
+        self._tracing_enabled = True
+        self._retraction_trace = []
+
+    def get_retraction_chain(self) -> List[Dict]:
+        """Return the accumulated retraction cascade trace.
+
+        Each entry is:
+            {trigger: str, retracted: [str], cascaded: [str], reason: str}
+        """
+        return list(self._retraction_trace)
 
     def explain_belief(self, belief_name: str) -> str:
         """Return a formatted explanation of a belief's justifications."""
--- a/tests/unit/argumentation_analysis/test_jtms_cascade_render.py
+++ b/tests/unit/argumentation_analysis/test_jtms_cascade_render.py
@@ -1,0 +1,214 @@
+"""Tests for JTMS retraction cascade rendering (#350).
+
+Validates that JTMS tracks cascading retractions when fallacies undermine
+beliefs, and that the cascade chain is properly stored in state.
+"""
+
+import asyncio
+import pytest
+
+from argumentation_analysis.services.jtms.jtms_core import JTMS, Belief
+from argumentation_analysis.orchestration.invoke_callables import _invoke_jtms
+from argumentation_analysis.core.shared_state import UnifiedAnalysisState
+from argumentation_analysis.orchestration.state_writers import _write_jtms_to_state
+
+
+def _run(coro):
+    return asyncio.get_event_loop().run_until_complete(coro)
+
+
+class TestJTMSCoreRetractionTrace:
+    """Tests for JTMS core retraction trace."""
+
+    def test_enable_tracing(self):
+        jtms = JTMS()
+        jtms.enable_tracing()
+        assert jtms._tracing_enabled is True
+        assert jtms._retraction_trace == []
+
+    def test_retraction_recorded_when_belief_set_false(self):
+        jtms = JTMS()
+        jtms.add_belief("A")
+        jtms.set_belief_validity("A", True)
+        jtms.enable_tracing()
+        jtms.set_belief_validity("A", False)
+        chain = jtms.get_retraction_chain()
+        assert len(chain) == 1
+        assert chain[0]["trigger"] == "A"
+        assert "A" in chain[0]["retracted"]
+
+    def test_cascade_retraction_tracked(self):
+        """A → B; retract A → cascade retracts B."""
+        jtms = JTMS()
+        jtms.add_belief("A")
+        jtms.add_belief("B")
+        jtms.add_justification(["A"], [], "B")
+        jtms.set_belief_validity("A", True)
+        # B should be valid via A
+        assert jtms.beliefs["B"].valid is True
+
+        jtms.enable_tracing()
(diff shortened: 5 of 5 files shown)


jsboigeEpita (commented):
# PR Review: #377, #378, #379

---

## PR #377: feat(demo): spectacular analysis demo script + Jupyter notebook (#361 #362)

**1131 lines added across 4 files**

**Verdict: REQUEST CHANGES**

### Summary

Adds a pedagogical demo script (`demonstration_epita_spectacular.py`) and a companion Jupyter notebook (`spectacular_analysis_tour.ipynb`) that showcase the analysis pipeline via hardcoded mock data. Includes two test files with subprocess-based smoke tests.

### Issues

**1. Hardcoded mock data -- no path to real pipeline integration**

The entire demo is built around `MOCK_SPECTACULAR_RESULT`, a 200+ line dict literal. The script's docstring says "When the real pipeline is available, replace MOCK_RUN_STEP with real calls" but there is no `MOCK_RUN_STEP` abstraction, no interface, and no wiring point. The script never calls any code from `argumentation_analysis/`. This means the demo will drift out of sync with the actual pipeline output format silently. A single mock data class or a protocol defining the expected shape would provide a migration path.

**2. Unicode box-drawing bug**

```python
print(_c(C.C, "╚" + "═" * w + "╗" if not subtitle else "╚" + "═" * w + "╝"))
```

Operator precedence: `if not subtitle` binds to the last `"═" * w`, not to the whole ternary. The ternary evaluates as `"╚" + ("═" * w + "╗" if not subtitle else "═" * w) + "╝"`. This means the bottom box always ends with `╝` regardless of subtitle presence, and the `╗`/`╝` branch is dead code. The result happens to be correct visually (the bottom corner should always be `╝`), but the code is misleading and the intent is unclear. It should be rewritten explicitly:

```python
bottom = "╝" if subtitle else "╗"  # or more likely, always "╝"
```

Actually, re-reading: for a box, the bottom-right corner should always be `╝`. The original code likely intended to use `╗` without subtitle and `╝` with subtitle, which is backwards. Either way, the operator precedence makes this confusing and worth fixing.

**3. `C.C` attribute shadowing**

```python
class C:
    C = "\033[96m"   # cyan
```

The class has a class attribute `C` where the class itself is named `C`. This works but is a readability hazard. Accessing `C.C` from inside or outside the class is fine, but it is unnecessarily confusing. Rename the class to `Colors` or rename the attribute to `CYAN`.

**4. Notebook headless execution test leaves artifacts on failure**

```python
try:
    ...
finally:
    executed.unlink(missing_ok=True)
```

The `try/finally` cleanup is good, but if the `nbconvert` call writes to a different filename or the test fails before `executed` is defined, `NameError` in the `finally` block would mask the original failure. More importantly, the test runs `nbconvert` with `--output` to a fixed path in the source tree (`spectacular_analysis_tour_executed.ipynb`). Under parallel test execution or if the test is interrupted, this file lingers. Use `tmp_path` fixture instead.

**5. Test subprocess calls with hardcoded script path**

The test resolves `SCRIPT` via relative path manipulation:
```python
SCRIPT = Path(__file__).resolve().parent.parent.parent.parent / "examples/..."
```

This is fragile. Four levels of `.parent` is easy to break if the test file moves. The `skipif(not SCRIPT.exists())` guard handles it gracefully, but the path construction itself should use a project root constant or `importlib` import.

**6. Test `_run` helper has unused `timeout` parameter**

```python
def _run(args=None, timeout=30):
    ...
    return subprocess.run(cmd, capture_output=True, text=True, timeout=timeout)
```

The `timeout` is passed to `subprocess.run` but never overridden in any test call. This is fine as a default, but it means a hung subprocess will kill the test with `subprocess.TimeoutExpired`. Worth documenting the intent.

**7. No French accents in generated text -- ASCII-only output strings**

The mock data and renderers use only ASCII text (e.g., "arnaque financée par George Soros" would have accents in proper French but the renderers use ASCII fallbacks). This is probably intentional for terminal compatibility, but worth a comment.

### Minor observations

- The test coverage is good: JSON output validation, step-specific content checks, invalid input handling, structural integrity.
- The `--json` flag and `--step` filter are useful CLI features.
- The notebook test checks for sensitive data leaks (`OPENAI_API_KEY`, etc.) -- good practice.

### Required changes before merge

1. Fix the box-drawing ternary operator precedence (issue 2).
2. Add at minimum a comment or a `TODO(#347)` with a concrete interface describing how the mock data will be replaced with real pipeline calls.
3. Rename `C.C` attribute or the `C` class to avoid shadowing (issue 3).
4. Use `tmp_path` in the notebook execution test instead of writing to the source tree (issue 4).

---

## PR #378: feat(jtms): retraction cascade rendering (#350)

**350 lines changed across 5 files**

**Verdict: APPROVE**

### Summary

Adds retraction cascade tracing to the JTMS (Justification-Based Truth Maintenance System). When a belief is retracted (set to False/None), the system now records which other beliefs cascade into invalidity as a result. The trace is propagated through the orchestration layer into `UnifiedAnalysisState`.

### What was done well

- **Minimal, focused change.** The feature is confined to `jtms_core.py` with clean wiring through `invoke_callables.py` and `state_writers.py`.
- **Opt-in tracing.** `enable_tracing()` must be called explicitly; no performance overhead when not in use.
- **Defensive copy.** `get_retraction_chain()` returns `list(self._retraction_trace)` rather than the internal reference.
- **Thorough tests.** 214 lines of tests covering: basic tracing, cascade detection, no-trace-when-disabled, deep cascades, format validation, copy semantics, orchestration integration, and state writer integration.

### Code-level observations

**1. `hasattr(self, '_jtms_ref')` check is fragile**

In `Belief.propagate()`:
```python
if (
    hasattr(self, '_jtms_ref')
    and self._jtms_ref is not None
    ...
):
```

The `_jtms_ref` attribute is set in `JTMS.add_belief()` but only after the PR's changes. Any `Belief` created outside of `add_belief()` (e.g., direct instantiation in tests or other code paths) will silently skip tracing. The `hasattr` guard handles this safely, but it would be cleaner to initialize `_jtms_ref = None` in `Belief.__init__` so the attribute always exists. This makes the contract explicit rather than relying on duck-typing.

**2. Cascade only appends to the last trace entry**

```python
self._jtms_ref._retraction_trace[-1]["cascaded"].append(conclusion.name)
```

This assumes the most recent retraction trace entry is the one whose cascade is currently being propagated. This is correct for single-belief retractions (which is the current usage pattern in `_invoke_jtms`), but if two beliefs are retracted in sequence without propagation completing between them, cascaded retractions from the first could be attributed to the second. The current code is safe because `set_belief_validity` calls `set_truth_value` which triggers `propagate` synchronously before returning, so by the time the next `set_belief_validity` runs, propagation is complete. But this coupling is implicit. A brief comment explaining the assumption would help future maintainers.

**3. `old_valid is True` vs `old_valid == True`**

Using `is True` identity checks is intentional and correct here (distinguishing `True` from truthy values like `1`), but `conclusion.valid is not True` also catches `None` and `False`. Since `valid` is typed as `Optional[bool]`, both `False` and `None` represent "not valid" states. This is semantically correct but worth noting: a belief transitioning from `True` to `None` (unknown) is recorded as a retraction. This may or may not be the desired behavior -- if "unknown" should not count as a retraction, add `and conclusion.valid is False` or `and validity is False`.

**4. `_run` helper uses deprecated event loop API**

```python
def _run(coro):
```

`asyncio.get_event_loop()` is deprecated in Python 3.10+ when no running loop exists. The existing codebase likely has this pattern already, so this is not a regression, but it is a tech debt item.

**5. Test for `_invoke_jtms` integration has weak assertions**

```python
def test_cascade_chain_with_fallacy(self):
    ...
    chain = result["retraction_chain"]
    assert len(chain) >= 1
    entry = chain[0]
    assert entry["trigger"]  # non-empty trigger
```

The test asserts `len(chain) >= 1` and that the trigger is truthy, but does not verify *which* belief triggered the cascade or *which* beliefs cascaded. This makes it possible for the test to pass even if the cascade logic is completely wrong (as long as *something* is retracted). Consider asserting that the trigger corresponds to the undermined argument's belief.

### No security issues

No user input flows unescaped. The tracing is purely internal state tracking.

### Verdict

Solid, well-tested feature addition. The observations above are minor improvements, none blocking. Approving with a recommendation to initialize `_jtms_ref = None` in `Belief.__init__` and strengthen the integration test assertions.

---

## PR #379: feat(synthesis): narrative synthesis agent (#351)

**481 lines changed across 5 files**

**Verdict: REQUEST CHANGES**

### Summary

Adds a template-based narrative synthesis plugin that weaves results from all analysis phases into 1-2 paragraphs of French prose. Includes a `NarrativeSynthesisPlugin` class for Semantic Kernel integration, an invoke callable, a state writer, and tests.

### Issues

**1. Silenced exceptions in `_invoke_narrative_synthesis`**

```python
for key, value in context.items():
    if key.startswith("phase_") and key.endswith("_output"):
        cap = key[len("phase_") : -len("_output")]
        writer = CAPABILITY_STATE_WRITERS.get(cap)
        if writer and isinstance(value, dict):
            try:
                writer(value, state, context)
            except Exception:
                pass
```

This bare `except Exception: pass` swallows all errors from state writers during the reconstruction loop. If a writer raises due to a bug or a schema change, the narrative will silently omit that data with no indication anything went wrong. At minimum, this should log the exception:

```python
except Exception:
    logger.warning("State writer %s failed for phase %s", cap, key, exc_info=True)
```

The `logging` import is already present in the file (used by other invoke callables).

**2. `_count_referenced_fields` uses hardcoded field list**

```python
for field_name in [
    "argument_quality_scores",
    "identified_fallacies",
    ...
]:
```

This list must be kept in sync with the fields checked in `build_narrative()`. If `build_narrative` is updated to reference a new field, `_count_referenced_fields` will undercount. If a field is removed from `build_narrative` but left in the count list, it will overcount. Consider extracting the field list as a shared constant, or computing the count from the same logic used by `build_narrative` (e.g., checking which `parts` were appended).

**3. Narrative template uses `getattr` instead of direct attribute access**

Throughout `build_narrative`:
```python
quality = getattr(state, "argument_quality_scores", {})
fallacies = getattr(state, "identified_fallacies", {})
```

This is defensive but unnecessary when the function's documented contract is that it receives a `UnifiedAnalysisState`. Using `getattr` hides type errors -- if the attribute is renamed or removed, the narrative silently omits that section rather than failing loudly. Since `UnifiedAnalysisState` is a concrete class with defined attributes, direct access (`state.argument_quality_scores`) would surface mismatches at development time. The `getattr` approach is appropriate for the `NarrativeSynthesisPlugin.synthesize()` method which accepts arbitrary dicts, but not for `build_narrative(state: Any)` when called from the pipeline with a real state object.

**4. `state: Any` type annotation**

```python
def build_narrative(state: Any) -> str:
```

Using `Any` defeats type checking. The function should accept `UnifiedAnalysisState` (or a `Protocol` if duck-typing is needed). The SK plugin wrapper can keep `Any` since it constructs a namespace object from a dict.

**5. Paragraph splitting logic is fragile**

```python
paragraph_1 = " ".join(parts[: len(parts) // 2 + 1])
paragraph_2 = " ".join(parts[len(parts) // 2 + 1 :]) if len(parts) > 3 else ""
```

With 7 parts (quality, fallacies, counters, JTMS, ATMS, Dung, formal logic), this produces paragraphs of 4 and 3 parts respectively. With 4 parts, it produces a single paragraph (3+1, but `len > 3` is true so paragraph_2 gets 0 parts joined = ""). Wait -- `len(parts) // 2 + 1` for 4 parts = 3, so paragraph_1 gets parts[0:3], paragraph_2 gets parts[3:] = 1 part. Since `len(parts) > 3` is true, paragraph_2 = " ".join([parts[3]]) which is non-empty. The logic works but the boundary condition at exactly 3 parts is off: `len(parts) // 2 + 1 = 2`, paragraph_1 = parts[0:2], paragraph_2 = parts[2:] = 1 part, but `len(parts) > 3` is False, so paragraph_2 = "". This means with 3 sections, you get a single paragraph containing 2 sections, and the 3rd section is silently dropped.

**This is a bug.** When exactly 3 analysis phases produce data, the third section's text is discarded.

**6. French text without accents (ASCII-only)**

The narrative template uses `"qualite"`, `"sophisme"`, `"croyance"`, `"detecte"`, `"etablit"`, etc. -- all without accents. This is presumably for ASCII-safe output, but it produces incorrect French text. If the system is used in a French academic context (EPITA), producing unaccented French looks unprofessional. Consider using proper UTF-8 French text (`"qualite" -> "qualite"`, `"sophisme" -> "sophisme"`, etc.) since the rest of the codebase already uses UTF-8 (the mock data in PR #377 has accented text).

**7. `paragraph_count` computation is wrong**

```python
"paragraph_count": narrative.count("\n\n") + 1,
```

If `narrative` is empty, this returns `1`. If `narrative` is `"\n\n"`, this returns `2`. The count is used as a metric but its definition is inconsistent with the actual paragraph count. For an empty narrative, 0 would be more accurate.

### Test quality observations

- Tests are well-structured with a `_rich_state()` factory that populates multiple fields.
- `test_at_least_5_distinct_references` is a good integration check.
- `test_readable_prose_not_json` is a creative way to ensure template output is prose, not structured data.
- Missing: no test for the `_count_referenced_fields` helper directly (it is only tested implicitly through the invoke callable).
- Missing: no test for the paragraph splitting edge case (3 parts) described in issue 5.

### No security issues

No user input flows into the narrative. The template uses only data from `UnifiedAnalysisState` fields populated by prior pipeline phases.

### Required changes before merge

1. Fix the silent `except Exception: pass` -- at minimum add logging (issue 1).
2. Fix the paragraph splitting bug at exactly 3 parts (issue 5).
3. Either use proper French accents or add a comment explaining the ASCII-only constraint (issue 6).
4. Consider typing `build_narrative(state)` more precisely than `Any` (issue 4).

### Recommended improvements

- Extract the shared field list between `build_narrative` and `_count_referenced_fields` (issue 2).
- Add a test for the 3-parts edge case in paragraph splitting.

Issue #364, opened 2026-04-25
[4.4] Rich CLI output formatter


jsboigeEpita:
**Parent Epic:** #344
**Track:** 4 — Démo Epita Escalation

## Goal
Pretty-print du state spectacular en CLI : sections, couleurs, citations inter-champs.

## Files to touch
- Nouveau `argumentation_analysis/cli/output_formatter.py`
- Wiring dans `run_orchestration.py` (option `--rich-output`)

## Definition of Done
- [ ] Sections : Extraction / Formal / Fallacies / JTMS / ATMS / Dung / Counter-arg / Debate / Quality / Narrative
- [ ] Couleurs (rich/colorama) et indentation lisible
- [ ] Citations : "voir Section X.Y" dans la narration
- [ ] Mode JSON brut conservé en parallèle (non breaking)
- [ ] Test `test_rich_output_formatter.py` passe
- [ ] PR mergée, CI green

## Dependencies
**Depends on:** #1.1 (state schema défini)
**Parallel with:** #4.1, #4.2, #4.3

## Assignee
po-2023

PR #380, opened 2026-04-30
feat(orchestration): conversational spectacular state upgrade (#363)

### LINKED ISSUE #363: [4.3] Conversational pipeline upgrade (spectacular state)
**Parent Epic:** #344
**Track:** 4 — Démo Epita Escalation

## Goal
Faire que `--mode conversational` du `run_orchestration.py` produise un output matching le workflow spectacular : PM ventile les capabilities, agents dialoguent, narration finale.

## Files to touch
- `argumentation_analysis/orchestration/conversational_orchestrator.py`
- Plugins ventilés via PM (selon CapabilityRegistry)

## Definition of Done
- [ ] `--mode conversational` sur doc_A produit state ≥ 28/32 champs
- [ ] Agents dialoguent visiblement (PM coordonne)
- [ ] Final state cohérent avec ce que produirait `--workflow spectacular`
- [ ] Test `test_conversational_spectacular.py` passe
- [ ] PR mergée, CI green

## Dependencies
**Depends on:** #1.1
**Parallel with:** #4.1, #4.2, #4.4

## Assignee
po-2023

jsboigeEpita:
## Summary

- Switch `run_conversational_analysis()` to use `UnifiedAnalysisState` (32 fields) when `spectacular=True` (new default)
- Result format now includes `capabilities_used`, `capabilities_missing`, `state_coverage_pct`, and unified pipeline summary — matching `--workflow spectacular` output
- Agent-to-capability mapping extracts from conversation log (ExtractAgent→fact_extraction, InformalAgent→fallacy_detection, etc.)
- Backward compatible: `spectacular=False` falls back to `RhetoricalAnalysisState`

## Test plan

- [x] `pytest tests/unit/argumentation_analysis/orchestration/test_conversational_spectacular.py` — 9/9 passed
- [x] State upgrade: spectacular=True → UnifiedAnalysisState, spectacular=False → RhetoricalAnalysisState
- [x] Result format: all unified pipeline keys present (workflow_name, summary, capabilities_used, state_snapshot)
- [x] Capability mapping: 7 capabilities extracted from mock conversation log
- [x] Field coverage: 28+/32 fields non-empty after mock population

Closes #363

### DIFF
--- a/argumentation_analysis/orchestration/conversational_orchestrator.py
+++ b/argumentation_analysis/orchestration/conversational_orchestrator.py
@@ -33,7 +33,10 @@ from semantic_kernel.connectors.ai.function_choice_behavior import (
 )
 from semantic_kernel.contents.chat_history import ChatHistory
 
-from argumentation_analysis.core.shared_state import RhetoricalAnalysisState
+from argumentation_analysis.core.shared_state import (
+    RhetoricalAnalysisState,
+    UnifiedAnalysisState,
+)
 from argumentation_analysis.core.state_manager_plugin import StateManagerPlugin
 from argumentation_analysis.orchestration.trace_analyzer import (
     ConversationalTraceAnalyzer,
@@ -301,6 +304,7 @@ async def run_conversational_analysis(
     text: str,
     max_turns_per_phase: int = 5,
     agent_names: Optional[List[str]] = None,
+    spectacular: bool = True,
 ) -> Dict[str, Any]:
     """Run a full conversational analysis on the input text.
 
@@ -308,6 +312,13 @@ async def run_conversational_analysis(
     1. Extraction + Detection (PM, ExtractAgent, InformalAgent)
     2. Formal Analysis (PM, FormalAgent, QualityAgent)
     3. Synthesis (PM, DebateAgent, CounterAgent, GovernanceAgent)
+
+    Args:
+        text: Input text to analyze.
+        max_turns_per_phase: Max agent turns per phase.
+        agent_names: Optional subset of agent names to use.
+        spectacular: If True, use UnifiedAnalysisState for 28+/32 field
+            coverage matching the spectacular workflow profile (#363).
 
     Returns dict with state snapshot, conversation history, and metrics.
     """
@@ -329,8 +340,9 @@ async def run_conversational_analysis(
     )
     kernel.add_service(llm_service)
 
-    # 2. Setup shared state
-    state = RhetoricalAnalysisState(text)
+    # 2. Setup shared state (#363: UnifiedAnalysisState for spectacular coverage)
+    state_cls = UnifiedAnalysisState if spectacular else RhetoricalAnalysisState
+    state = state_cls(text)
 
     # 3. Create all agents
     all_agents = create_conversational_agents(
@@ -543,15 +555,54 @@ async def run_conversational_analysis(
 
     result = {
         "mode": "conversational",
+        "workflow_name": "spectacular_analysis" if spectacular else "conversational",
         "phases": [p["name"] for p in phase_configs],
         "conversation_log": conversation_log,
         "total_messages": len(conversation_log),
         "duration_seconds": duration,
         "state_snapshot": state_snapshot,
         "state_non_empty_fields": non_empty,
+        "state_total_fields": len(state_snapshot),
+        "state_coverage_pct": (
+            non_empty / len(state_snapshot) * 100 if state_snapshot else 0
+        ),
         "unified_state": state,
         "trace_report": trace_report,
+        "summary": {
+            "completed": len(phase_configs),
+            "failed": 0,
+            "skipped": 0,
+            "total": len(phase_configs),
+            "total_messages": len(conversation_log),
+        },
     }
+
+    # Spectacular mode: add capability mapping from conversation log
+    if spectacular:
+        capabilities_used = set()
+        for msg in conversation_log:
+            agent = msg.get("agent", "")
+            if agent == "ExtractAgent":
+                capabilities_used.add("fact_extraction")
+            elif agent == "InformalAgent":
+                capabilities_used.update(
+                    ["neural_fallacy_detection", "hierarchical_fallacy_detection"]
+                )
+            elif agent == "FormalAgent":
+                capabilities_used.update(
+                    ["nl_to_logic_translation", "fol_reasoning", "modal_logic",
+                     "propositional_logic"]
+                )
+            elif agent == "QualityAgent":
+                capabilities_used.add("argument_quality")
+            elif agent == "CounterAgent":
+                capabilities_used.add("counter_argument_generation")
+            elif agent == "DebateAgent":
+                capabilities_used.add("adversarial_debate")
+            elif agent == "GovernanceAgent":
+                capabilities_used.add("governance_simulation")
+        result["capabilities_used"] = list(capabilities_used)
+        result["capabilities_missing"] = []
 
     logger.info(
         f"Conversational analysis complete: {len(conversation_log)} messages, "
--- a/argumentation_analysis/orchestration/unified_pipeline.py
+++ b/argumentation_analysis/orchestration/unified_pipeline.py
@@ -119,21 +119,9 @@ async def run_unified_analysis(
                 run_conversational_analysis,
             )
 
-            conv_result = await run_conversational_analysis(text)
+            conv_result = await run_conversational_analysis(text, spectacular=True)
 
-            # Normalize result format for benchmark compatibility (#208-L)
-            # Conversational returns {phases: [names], conversation_log, total_messages}
-            # Benchmark expects {phases: {dict}, summary: {completed, total, ...}}
-            phase_names = conv_result.get("phases", [])
-            total_msgs = conv_result.get("total_messages", 0)
-            conv_result["summary"] = {
-                "completed": len(phase_names),
-                "failed": 0,
-                "skipped": 0,
-                "total": len(phase_names),
-                "total_messages": total_msgs,
-            }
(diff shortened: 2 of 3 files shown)


jsboigeEpita (commented):
# PR Review: #380, #381, #382

---

## PR #380 — feat(orchestration): conversational spectacular state upgrade (#363)

**Verdict: APPROVE with minor comments**

### Summary

Adds a `spectacular=True` parameter to `run_conversational_analysis` that switches from `RhetoricalAnalysisState` to `UnifiedAnalysisState`, enriching the result dict with coverage metrics, capability mapping, and a normalized `summary` block. The unified pipeline call site is updated to pass `spectacular=True` and the normalization logic it previously duplicated is removed.

### Observations

**Correctness — Good**

- The state class switch (`UnifiedAnalysisState if spectacular else RhetoricalAnalysisState`) is clean and backward-compatible. Existing callers that omit the parameter default to `spectacular=True`, which is the intended new default.
- The `summary` dict (completed/failed/skipped/total) is now produced once inside `run_conversational_analysis` rather than being reassembled by the caller. This is the right move for consistency.
- Coverage calculation (`non_empty / len(state_snapshot) * 100`) has a guard for empty snapshots.

**Capability mapping — acceptable but fragile**

- Lines 79-101 in `conversational_orchestrator.py` map agent names to capability strings via a chain of `if/elif` on string literals. This duplicates knowledge already present in the agent definitions themselves. If a new agent is added or renamed, this block must be manually updated or it silently drops the capability.
- `capabilities_missing` is hardcoded to `[]`. This is acknowledged by the field existing at all (schema completeness), but it means downstream consumers that rely on it for gap analysis will always see a clean slate. Consider at minimum a docstring noting this is a placeholder.

**Test quality — Good**

- The `mock_conversational_deps` fixture properly patches `sk`, `ConversationalTraceAnalyzer`, `create_conversational_agents`, and env vars.
- `test_spectacular_false_uses_rhetorical_state` correctly verifies the negative case and uses `not isinstance` to confirm it is not a `UnifiedAnalysisState`.
- `test_capabilities_extracted_from_conversation` provides a realistic conversation log with all 7 agent types and verifies each expected capability.
- `test_state_populates_to_28_plus` is a thorough integration-style test that manually populates 28+ fields and verifies coverage calculation.

**Minor issues**

1. The `inspect` import at the top of the test file is unused — should be removed.
2. The `spectacular` parameter defaults to `True`. This is a behavioral change for any caller relying on the old default (which implicitly got `RhetoricalAnalysisState`). The PR description frames this as intentional, but it should be noted in a changelog entry.

---

## PR #381 — feat(cli): rich output formatter for spectacular analysis (#364)

**Verdict: APPROVE with minor comments**

### Summary

Introduces a `cli/output_formatter.py` module that renders `UnifiedAnalysisState` snapshots into 10 numbered, cross-referenced sections using Rich (with a plain-text fallback). Integrates into `run_orchestration.py` via a `--rich-output` CLI flag for both `pipeline` and `conversational` modes.

### Observations

**Correctness — Good**

- The `HAS_RICH` import guard and `_render_plain` fallback ensure the module works without Rich installed. The test for this path (`test_fallback_to_plain_when_no_console`) confirms it.
- Section numbering is consistent: the `SECTIONS` list defines 1-10, and each `_render_*` function uses the correct number. Section 5 (ATMS) has no dedicated renderer — this is fine since the ATMS data is typically empty in current outputs, but the gap should be documented or a minimal renderer added for completeness.

**Test quality — Good**

- Fixtures are well-designed: `sample_result` has all 10 sections populated, `empty_result` tests the edge case.
- Individual section renderer tests (`TestSectionRenderers`) cover each `_render_*` function with both populated and empty state.
- `TestCrossReferences` verifies the cross-ref system works (quality refs fallacies).
- The `test_rich_header_content` test verifies the first call produces a Rich Panel by checking `hasattr(first_call_arg, "title")` — a reasonable approach given we're working with MagicMock.

**Consistency — One concern**

- The conversational mode integration (lines 382-400 in `run_orchestration.py`) builds a `render_result` dict by manually extracting and normalizing keys from the conversational result. This is a second normalization site — the PR #380 changes moved normalization *into* `run_conversational_analysis`, so this block should now be unnecessary if the conversational result already has the right shape. If the two PRs are meant to land together, this duplication should be cleaned up. If #381 lands first, it's acceptable as temporary scaffolding.

**Minor issues**

1. The `__init__.py` file in `cli/` is empty — fine for now, but if the module grows, a proper `__all__` export list would be helpful.
2. `_render_quality` (line 268) always prints a cross-reference to Section 3 even when there are no fallacies. The cross-ref line is unconditional within the loop over `argument_quality_scores`, which could mislead users if fallacy detection was not run.
3. `_truncate` does not handle `None` input — it would crash on `_truncate(None)`. The callers generally pass `str(...)` wrapped values, but a defensive `if text is None: return ""` would be safer.
4. `json` is imported in the test file but never used.

---

## PR #382 — feat(orchestration): modern Sherlock multi-agent orchestrator (#357)

**Verdict: REQUEST CHANGES**

### Summary

Introduces `SherlockModernOrchestrator`, an investigation-themed orchestrator that wires 7 agent callables sequentially through 7 phases (extraction, fallacy detection, quality, cross-examination, belief tracking, hypothesis branching, solution synthesis). Registers in the workflow catalog and adds a `sherlock_modern` CLI mode.

### Critical Issues

**1. `_run()` helper uses deprecated event loop API**

```python
def _run(coro):
```

`asyncio.get_event_loop()` emits a `DeprecationWarning` since Python 3.10 and will behave differently in Python 3.12+ when no running loop exists. The tests already use `@pytest.mark.asyncio` on the PR #380 tests (the correct pattern). This test file should do the same. Every test in `TestSherlockModernOrchestrator` uses `_run()` — 13 call sites.

**Recommended fix:** Add `@pytest.mark.asyncio` to all async tests and `await` the coroutines directly. Remove the `_run` helper entirely.

**2. `_build_result` silently swallows state snapshot errors**

```python
def _build_result(self) -> InvestigationResult:
    snapshot = None
    try:
        snapshot = self.state.get_state_snapshot(summarize=True)
    except Exception:
        pass
```

A bare `except Exception: pass` hides bugs. If `self.state` is `None` or `get_state_snapshot` raises due to a real issue (corrupted state, missing attribute), the caller gets `None` with no indication of what went wrong. At minimum, log the exception:

```python
except Exception as e:
    logger.warning("Failed to build state snapshot: %s", e)
```

**3. `build_sherlock_modern_workflow` silently returns `None` on any import or build error**

```python
def build_sherlock_modern_workflow():
    try:
        ...
        return WorkflowBuilder(...).build()
    except Exception:
        return None
```

This follows the existing pattern in `workflows.py` (other builders do the same), so it's consistent with the codebase. However, the registration site in `workflows.py` then does:

```python
WORKFLOW_CATALOG["sherlock_modern"] = build_sherlock_modern_workflow()
```

If the builder returns `None`, the catalog entry maps to `None`. Any code that does `catalog["sherlock_modern"].phases` will crash with `AttributeError`. The existing codebase has this problem for other workflows too, but this PR should not propagate it. Add a guard:

```python
wf = build_sherlock_modern_workflow()
if wf is not None:
    WORKFLOW_CATALOG["sherlock_modern"] = wf
```

**4. Missing Section 5 (ATMS) renderer in PR #381 is partially caused by this PR**

The `investigate()` method runs `_phase_hypothesis_branching` which invokes ATMS, but the results go into `self._hypotheses` and the trace — they are not written to `self.state` at all. The `state_snapshot` in the final result will never contain ATMS data from this orchestrator. If the Sherlock orchestrator is meant to populate `UnifiedAnalysisState` (as its constructor accepts one), each phase should write findings back to the state object, not just to internal trace structures.

**5. `investigate()` does not populate `self.state`**

Looking at all 7 phase methods: none of them write to `self.state`. They only append to `self._trace` and store results in `self._ctx`. The `UnifiedAnalysisState` passed in (or created) remains empty except for `raw_text`. This means:
- The `state_snapshot` in the result will show 0 coverage.
- The rich formatter from PR #381 will render an empty dashboard for sherlock_modern results.
- The `spectacular` coverage promise (28+ fields) is unmet.

Each phase method should be populating the corresponding state fields (e.g., `_phase_extraction` should set `self.state.identified_arguments`, `_phase_fallacy_detection` should set `self.state.identified_fallacies`, etc.).

### Test Quality — Needs improvement

- Tests only verify structural properties (agent count, trace length, field presence) but never verify that the state is actually populated with meaningful data.
- `test_no_llm_calls_template_fallback` is the right idea but only verifies that `_invoke_safe` returns fallback dicts — it doesn't verify the orchestrator correctly interprets those fallbacks into the trace.
- No test verifies the `state_snapshot` content after investigation.
- No test for the `context` parameter beyond "doesn't crash."
- `TestInvestigationResult` and `TestInvestigationStep` test dataclass defaults — low value, not harmful but not informative.

### Positive Observations

- The `InvestigationStep` and `InvestigationResult` dataclasses are clean and well-structured.
- The `_invoke_safe` pattern with fallback dicts is a good resilience pattern for when LLM services are unavailable.
- The investigation metaphor (claims/inconsistencies/cross-examination) is consistently applied throughout.
- CLI integration is clean and follows existing patterns in `run_orchestration.py`.
- The workflow DSL registration follows the established pattern.

### Summary of required changes

| Priority | Issue | Location |
|----------|-------|----------|
| P0 | Phase methods must populate `self.state` fields | All 7 `_phase_*` methods |
| P1 | Replace deprecated `_run()` with `@pytest.mark.asyncio` | Test file, 13 call sites |
| P1 | Guard against `None` workflow in catalog registration | `workflows.py` |
| P2 | Log exception in `_build_result` instead of bare `pass` | `sherlock_modern_orchestrator.py` |

PR #385, opened 2026-04-30
feat(scenarios): Sherlock Modern investigation scenarios with opaque IDs (#360)

jsboigeEpita:
## Summary
- 3 investigation scenarios (`doc_A`, `doc_B`, `doc_C`) on encrypted dataset with opaque IDs
- Scenario runner with encrypted dataset loader + fallback for demo mode
- Privacy compliance tests ensure no plaintext content in commits

## Test plan
- [x] `pytest tests/unit/argumentation_analysis/test_sherlock_scenarios.py -v` — 7/7 passed
- [ ] Verify scenario runner works with real encrypted dataset
- [ ] Confirm no plaintext leaks in git diff

### DIFF
--- a/argumentation_analysis/agents/core/oracle/hypothesis_tracker.py
+++ b/argumentation_analysis/agents/core/oracle/hypothesis_tracker.py
@@ -1,0 +1,207 @@
+"""ATMS Hypothesis Tracker for Sherlock Modern (#359).
+
+Tracks >=3 simultaneous hypotheses via ATMS, updating coherence
+as new evidence arrives. Contradicted hypotheses trigger visible
+retractions with cascade tracking.
+
+Key API:
+- create_hypothesis() — declare a named hypothesis with assumptions
+- add_evidence() — feed new evidence, update all hypotheses
+- get_active_hypotheses() — list coherent hypotheses
+- get_retracted_hypotheses() — list contradicted hypotheses with reasons
+- get_investigation_state() — full snapshot for trace/reporting
+"""
+
+import logging
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional, Set, FrozenSet
+
+from argumentation_analysis.services.jtms.atms_core import ATMS, ATMSNode
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class Hypothesis:
+    """A tracked hypothesis with ATMS label."""
+
+    id: str
+    name: str
+    assumptions: List[str]
+    description: str = ""
+    coherent: bool = True
+    retraction_reason: str = ""
+    evidence_applied: List[str] = field(default_factory=list)
+    derived_nodes: Set[str] = field(default_factory=set)
+
+
+class HypothesisTracker:
+    """Tracks >=3 simultaneous hypotheses via ATMS.
+
+    Each hypothesis is a set of assumptions. The ATMS computes
+    minimal environments under which nodes can be derived.
+    Contradictions invalidate hypotheses.
+
+    Usage:
+        tracker = HypothesisTracker()
+        tracker.create_hypothesis("h_trust", "Full trust", ["source_reliable", "method_sound"])
+        tracker.create_hypothesis("h_doubt", "Skeptical", ["source_unreliable"])
+        tracker.add_evidence("claim_A", supports=["source_reliable"], contradicts=[])
+        active = tracker.get_active_hypotheses()
+    """
+
+    MIN_HYPOTHESES = 3
+
+    def __init__(self):
+        self._atms = ATMS()
+        self._hypotheses: Dict[str, Hypothesis] = {}
+        self._evidence_log: List[Dict[str, Any]] = []
+        self._assumption_to_hypothesis: Dict[str, str] = {}
+        self._contradicted_assumptions: Set[str] = set()
+
+    def create_hypothesis(
+        self,
+        hyp_id: str,
+        name: str,
+        assumptions: List[str],
+        description: str = "",
+    ) -> Hypothesis:
+        """Create a new hypothesis with ATMS assumptions."""
+        # Register assumptions in ATMS
+        atms_assumptions = []
+        for a in assumptions:
+            self._atms.add_assumption(a)
+            self._assumption_to_hypothesis[a] = hyp_id
+            atms_assumptions.append(a)
+
+        hyp = Hypothesis(
+            id=hyp_id,
+            name=name,
+            assumptions=list(assumptions),
+            description=description,
+        )
+        self._hypotheses[hyp_id] = hyp
+        return hyp
+
+    def add_evidence(
+        self,
+        evidence_id: str,
+        supports: Optional[List[str]] = None,
+        contradicts: Optional[List[str]] = None,
+        derives: Optional[List[str]] = None,
+    ) -> Dict[str, Any]:
+        """Add new evidence and update all hypotheses.
+
+        Args:
+            evidence_id: Unique name for this evidence.
+            supports: Assumption names this evidence supports.
+            contradicts: Assumption names this evidence contradicts.
+            derives: Additional nodes this evidence derives.
+
+        Returns:
+            Dict with updated hypothesis states.
+        """
+        supports = supports or []
+        contradicts = contradicts or []
+        derives = derives or []
+
+        # Register evidence node as assumption (evidence is an input fact,
+        # so it always has its own environment)
+        self._atms.add_assumption(evidence_id)
+
+        # Add justifications: supported assumptions → evidence node
+        if supports:
+            self._atms.add_justification(
+                in_names=supports,
+                out_names=[],
+                conclusion_name=evidence_id,
+            )
+
+        # Contradictions: evidence contradicts assumptions
+        # Track directly since ATMS invalidate_environment clears "⊥" label
+        if contradicts:
+            self._contradicted_assumptions.update(contradicts)
+            for c in contradicts:
+                if c not in self._atms.nodes:
+                    self._atms.add_node(c, is_assumption=False)
+
+        # Additional derivations
+        for d in derives:
+            self._atms.add_node(d, is_assumption=False)
+            self._atms.add_justification(
+                in_names=[evidence_id],
+                out_names=[],
+                conclusion_name=d,
+            )
+
+        # Log evidence
+        evidence_record = {
+            "evidence_id": evidence_id,
+            "supports": supports,
+            "contradicts": contradicts,
+            "derives": derives,
+        }
+        self._evidence_log.append(evidence_record)
+
+        # Update all hypotheses
+        self._update_hypotheses(evidence_id)
+
+        return {
+            "evidence_added": evidence_id,
+            "active_count": len(self.get_active_hypotheses()),
+            "retracted_count": len(self.get_retracted_hypotheses()),
+        }
+
+    def _update_hypotheses(self, new_evidence: str) -> None:
+        """Update hypothesis coherence based on contradicted assumptions."""
+        for hyp_id, hyp in self._hypotheses.items():
+            if not hyp.coherent:
+                continue
+
+            hyp.evidence_applied.append(new_evidence)
+
+            contradicted = set(hyp.assumptions) & self._contradicted_assumptions
+            if contradicted:
+                hyp.coherent = False
+                hyp.retraction_reason = (
+                    f"Contradicted by evidence '{new_evidence}' — "
+                    f"assumptions {contradicted} conflict with evidence"
+                )
+
+    def get_active_hypotheses(self) -> List[Hypothesis]:
+        """Get all currently coherent hypotheses."""
+        return [h for h in self._hypotheses.values() if h.coherent]
+
+    def get_retracted_hypotheses(self) -> List[Hypothesis]:
+        """Get all contradicted/retracted hypotheses."""
+        return [h for h in self._hypotheses.values() if not h.coherent]
+
+    def get_all_hypotheses(self) -> List[Hypothesis]:
+        """Get all hypotheses."""
+        return list(self._hypotheses.values())
+
+    def get_investigation_state(self) -> Dict[str, Any]:
+        """Full snapshot for investigation trace."""
+        return {
+            "total_hypotheses": len(self._hypotheses),
+            "active": [h.id for h in self.get_active_hypotheses()],
+            "retracted": [
+                {"id": h.id, "reason": h.retraction_reason}
+                for h in self.get_retracted_hypotheses()
+            ],
+            "evidence_count": len(self._evidence_log),
+            "evidence_log": list(self._evidence_log),
+        }
+
+    def get_hypothesis_comparison(self) -> str:
+        """Human-readable comparison of all hypotheses."""
+        lines = ["Hypothesis Comparison:"]
+        for hyp in self._hypotheses.values():
+            status = "COHERENT" if hyp.coherent else "RETRACTED"
+            lines.append(
+                f"  [{status}] {hyp.id} ({hyp.name}): "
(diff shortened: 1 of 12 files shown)


jsboigeEpita (commented):
# PR Review: #383, #384, #385 (Sherlock Modern Feature Suite)

---

## PR #383 — feat(investigation): open-domain whodunit-style argument analysis (#358)

**Verdict: REQUEST CHANGES**

### Files Reviewed

| File | Lines | Role |
|------|-------|------|
| `orchestration/sherlock_modern_orchestrator.py` | 388 (new) | Core orchestrator |
| `orchestration/open_domain_investigation.py` | 233 (new) | Attribution layer |
| `orchestration/workflows.py` | +9 | Catalog registration |
| `run_orchestration.py` | +36 | CLI mode |
| `demo_sherlock_investigation.py` | 80 (new) | Demo script |
| `test_sherlock_modern_orchestrator.py` | 208 (new) | Orchestrator tests |
| `test_open_domain_investigation.py` | 280 (new) | Investigation tests |

### Correctness Issues

1. **Orchestrator is not idempotent / not reusable.** `SherlockModernOrchestrator.investigate()` appends to `self._trace`, `self._agents`, and `self._hypotheses` but never resets them. Calling `investigate()` twice on the same instance will accumulate duplicate trace entries, double-count agents, and merge hypotheses from both runs. The `_build_result()` method will return corrupted data on the second call. Either add a `_reset()` call at the top of `investigate()`, or document that instances are single-use.

2. **`_invoke_safe` swallows all exceptions silently.** The method catches bare `Exception` and logs at `DEBUG` level. If `invoke_callables._invoke_extract` raises a `TypeError` (wrong signature) or an `ImportError` (module missing), the orchestrator silently falls through to empty fallbacks. Every phase will report success with zero findings. At minimum, log at `WARNING` level so operators can see that real callables are failing.

3. **`build_sherlock_modern_workflow()` silently returns `None` on any exception.** The bare `except Exception` block in `build_sherlock_modern_workflow` (line 633 of the diff) means if the `WorkflowBuilder` API changes or a phase name is invalid, the function returns `None` and the caller in `get_workflow_catalog` registers nothing -- silently. The catalog registration also catches exceptions, so the failure is double-swallowed.

4. **`_extract_claims_from_trace` always appends a quality score line, even with zero evaluated arguments.** Line 135: `claims.append(f"Reliability score: {score:.1f}/10 ({n} arguments)")` runs unconditionally for any `quality_evaluation` phase step, even when `n == 0`. This produces a misleading claim like "Reliability score: 0.0/10 (0 arguments)" which pollutes the attribution output.

5. **Confidence calculation is arbitrary and not evidence-based.** In `_build_attributions` (line 155), confidence for coherent hypotheses is `0.7 + 0.1 * min(len(assumptions), 3)`. This means a hypothesis with 3 assumptions gets confidence 1.0 and one with 4+ assumptions also gets 1.0. The formula has no grounding in the actual evidence -- it is purely structural. This should at minimum be documented as a placeholder, or better yet, derived from the ATMS label environments.

6. **Demo exit condition is trivially always true.** `sys.exit(0 if result.claims_analyzed >= 0 else 1)` -- `claims_analyzed` is a `len()` result, so it is always >= 0. This exit code always returns 0, making it useless as a success/failure signal.

### Test Quality Issues

7. **All orchestrator tests run without mocking `_invoke_safe`.** Tests like `test_investigate_returns_result` and `test_uses_at_least_5_agents` call `orch.investigate(SAMPLE_DISCOURSE)` without patching `_invoke_safe`. This means they actually try to import and call `invoke_callables._invoke_extract`, etc. These tests will either (a) fail if the invoke module is not available, or (b) accidentally exercise real codepaths. The only test that properly mocks `_invoke_safe` is `test_no_llm_calls_template_fallback`. Every other test should also mock it, or at minimum use the same pattern.

8. **`test_demo_has_run_function` is trivially true.** It creates a module spec and asserts `hasattr(mod, "__file__")`. Every Python module has `__file__`. This test verifies nothing about the demo's `run_demo()` function.

9. **Test helper `_run` uses deprecated `get_event_loop()`.** `asyncio.get_event_loop().run_until_complete()` is deprecated since Python 3.10 and emits `DeprecationWarning` in Python 3.12+. Should use `asyncio.run()` or `pytest-asyncio`.

10. **Massive patch-string duplication.** The full patch path `"argumentation_analysis.orchestration.sherlock_modern_orchestrator.SherlockModernOrchestrator.investigate"` is repeated 14 times across the investigation tests. Extract it into a constant or fixture.

### Consistency Issues

11. **`InvestigationStep` dataclass is defined but only used internally as an intermediate, then serialized to `dict` in `_build_result()`.** The trace is stored as `List[InvestigationStep]` but returned as `List[Dict]`. This is inconsistent -- either return `List[InvestigationStep]` in the result, or skip the dataclass entirely and use dicts throughout.

12. **`_phase_solution_synthesis` sets `self._solution` as an ad-hoc attribute** (line 535). All other state is stored in `self._trace`, `self._agents`, `self._hypotheses`, `self._ctx`. This breaks the pattern and `_solution` is not initialized in `__init__`, so accessing it before synthesis would raise `AttributeError`. The `getattr(self, "_solution", "")` in `_build_result` is a workaround for this.

### Security

13. **No security issues found.** The code does not handle user input in unsafe ways, does not write to the filesystem from untrusted data (the output path in CLI mode comes from `args.output` which is a file path argument), and the opaque ID pattern is correctly applied.

### Required Changes

- [ ] Reset internal state at the beginning of `investigate()` to make the orchestrator reusable
- [ ] Fix `_extract_claims_from_trace` to skip quality score line when `n == 0`
- [ ] Mock `_invoke_safe` in all orchestrator tests that don't currently do so
- [ ] Fix the demo exit condition to be meaningful (e.g., `>= 1`)
- [ ] Initialize `self._solution = ""` in `__init__`
- [ ] Remove or make meaningful `test_demo_has_run_function`

---

## PR #384 — feat(atms): hypothesis branching with evidence-driven retraction (#359)

**Verdict: REQUEST CHANGES**

### Files Reviewed

| File | Lines | Role |
|------|-------|------|
| `agents/core/oracle/hypothesis_tracker.py` | 207 (new) | ATMS hypothesis tracker |
| `test_sherlock_atms_branching.py` | 178 (new) | Tracker tests |

### Correctness Issues

1. **Contradiction tracking is done outside the ATMS, bypassing the core.** The `_contradicted_assumptions` set (line 66) is maintained manually. The ATMS has its own contradiction mechanism (`invalidate_environment`, nogood detection), but `_update_hypotheses` never queries the ATMS for its contradiction state. This means the ATMS label propagation and the HypothesisTracker's coherence state can diverge. For example, if the ATMS detects a nogood through its own inference rules (not through explicit `contradicts=` in `add_evidence`), the tracker will not notice.

2. **`_assumption_to_hypothesis` mapping silently overwrites on collision.** If two hypotheses share an assumption name (which is a valid scenario -- e.g., both "h_trust" and "h_conservative" might include "evidence_available"), `self._assumption_to_hypothesis[a] = hyp_id` will map the assumption to whichever hypothesis was created last. This mapping is never read back in the current code, but its presence suggests it was intended for something. Either remove it or document the collision behavior.

3. **Evidence added as both assumption and derived node creates circular ATMS structure.** In `add_evidence` (lines 116-141), the evidence ID is first registered as an assumption (`self._atms.add_assumption(evidence_id)`), then a justification is added where supported assumptions derive the evidence node (`add_justification(in_names=supports, conclusion_name=evidence_id)`). Since `evidence_id` is already an assumption, this justification is circular: the evidence node is assumed true AND derived from other assumptions. Depending on ATMS implementation, this may be ignored or may create spurious environments.

4. **`add_justification` `out_names` parameter is always empty `[]`.** Looking at the ATMS API pattern, justifications typically have `in_names` (antecedents) and `out_names` (consequents that become false if all antecedents hold). Passing empty `out_names` means the justification only creates a derivation from `in_names` to `conclusion_name` without any nogood registration. This may be intentional but contradicts standard ATMS usage where contradictions are expressed via out-names.

5. **MIN_HYPOTHESES = 3 is declared but never enforced.** The class attribute `MIN_HYPOTHESES = 3` is defined but `create_hypothesis` does not check whether enough hypotheses exist before allowing `get_investigation_state` or `get_active_hypotheses` to be called. The tests verify `>= 3` manually, but the class itself does not enforce this invariant.

6. **`_update_hypotheses` marks ALL hypotheses with contradicted assumptions as incoherent, but only blames the latest evidence.** Line 173: `f"Contradicted by evidence '{new_evidence}'"` is misleading when the hypothesis was actually contradicted by an earlier evidence addition. The retraction reason always names the most recent evidence, not the one that actually caused the contradiction.

### Test Quality Issues

7. **Tests never call the actual ATMS core.** `HypothesisTracker` depends on `argumentation_analysis.services.jtms.atms_core.ATMS` but no test verifies that the ATMS is correctly populated. Tests create hypotheses and add evidence, then check the tracker's own bookkeeping (`_contradicted_assumptions`, `coherent` flag) -- but never verify that `self._atms.nodes` contains the expected nodes, or that ATMS environments are correct. The ATMS integration is effectively untested.

8. **`test_retracted_hypothesis_has_reason` has a conditional assertion.** Line 921: `if retracted:` followed by `assert retracted[0].retraction_reason != ""`. If the list is empty, the test passes without checking anything. This should be `assert retracted, "Expected h1 to be retracted"` before checking the reason.

9. **No test for shared assumptions between hypotheses.** As noted in issue #2, if two hypotheses share an assumption and that assumption is contradicted, both should retract. No test covers this case.

10. **No test for evidence with both `supports` and `contradicts` simultaneously.** The `add_evidence` API accepts both parameters, but no test exercises them together.

11. **No test for the `derives` parameter.** The `derives` parameter in `add_evidence` is never tested.

### Consistency Issues

12. **Uses `Set` and `FrozenSet` from typing but only `Set` is actually used.** `FrozenSet` is imported but never referenced. Minor cleanup needed.

13. **`Hypothesis.evidence_applied` grows monotonically.** Every call to `add_evidence` appends to `evidence_applied` for all non-retracted hypotheses, even hypotheses that are unrelated to the evidence. After 100 evidence additions, every hypothesis will have 100 entries in `evidence_applied`, most of which are irrelevant. This should either track only relevant evidence or be a set.

### Security

14. **No security issues.** The code operates purely on in-memory data structures with no I/O.

### Required Changes

- [ ] Fix the circular ATMS structure: evidence should not be both an assumption and derived
- [ ] Fix retraction reason to identify the actual contradicting evidence, not just the latest
- [ ] Add a test for shared assumptions between hypotheses
- [ ] Fix `test_retracted_hypothesis_has_reason` to fail if no retraction occurs
- [ ] Remove unused `FrozenSet` import
- [ ] Either remove `_assumption_to_hypothesis` or document the overwrite behavior
- [ ] Add tests for `derives` parameter and for `supports` + `contradicts` together

---

## PR #385 — feat(scenarios): Sherlock Modern investigation scenarios with opaque IDs (#360)

**Verdict: APPROVE** (with minor suggestions)

### Files Reviewed

| File | Lines | Role |
|------|-------|------|
| `docs/sherlock/scenarios.md` | 83 (new) | Scenario documentation |
| `run_scenarios.py` | 140 (new) | Scenario runner |
| `test_sherlock_scenarios.py` | 146 (new) | Scenario tests |

### Correctness Issues

1. **`_load_dataset_text` uses `hash(doc_id)` for index selection, which is non-deterministic across Python runs.** Python's `hash()` function is randomized by default (PYTHONHASHSEED). This means `doc_A` might map to index 0 in one run and index 3 in another, producing different texts for the same document ID across executions. For a scenario runner, this undermines reproducibility. Use a deterministic hash (e.g., `hashlib.md5(doc_id.encode()).hexdigest()`) or a fixed mapping.

2. **`_load_dataset_text` silently returns a hardcoded fallback string on any exception.** The bare `except Exception: pass` on line 161 means if the encrypted config is corrupted, the key is wrong, or the definitions format changed, the runner will silently use the fallback text for all three scenarios. Every scenario will analyze the same text. At minimum, log a warning.

3. **`run_scenario` imports `OpenDomainInvestigator` inside the function but mocks `SherlockModernOrchestrator` in the test.** The mock patches `sherlock_modern_orchestrator.SherlockModernOrchestrator.investigate`, but `run_scenario` imports `OpenDomainInvestigator` which internally imports `SherlockModernOrchestrator`. Since `OpenDomainInvestigator.investigate_document` does a late import of `SherlockModernOrchestrator` (line 292 of PR #383), the mock should work -- but this is fragile and depends on the import happening after the mock is applied. Worth noting as a coupling risk.

### Test Quality Issues

4. **`test_opaque_ids_no_plaintext` has a confusing assertion.** Line 336: `assert pattern not in scenario_text.lower() or pattern == "author"`. This allows the string "author" to appear in scenario text, but "author" appears in the mock data in `test_sherlock_atms_branching.py` (PR #384), not in the scenario definitions. The assertion logic is correct for what it checks, but the `or pattern == "author"` exception is unexplained and makes the test weaker. Add a comment explaining why "author" is an exception.

5. **`test_run_scenario_returns_result` mocks the wrong path.** The test patches `sherlock_modern_orchestrator.SherlockModernOrchestrator.investigate`, but `run_scenario` calls `OpenDomainInvestigator().investigate_document()` which does a late import. The mock should target `open_domain_investigation.OpenDomainInvestigator` or patch at the right module level. Currently the mock may not actually intercept the call depending on import ordering.

6. **Privacy tests only check French phrases.** `test_no_plaintext_in_docs` checks for `"discours complet"` and `"texte integral"` (without accents) but the actual sensitive content in the encrypted dataset likely contains accented French (`"discours complet"`, `"texte integral"`). The assertion would miss accented variants.

### Consistency Issues

7. **Scenario runner uses a different sys.path manipulation than the demo.** Both `run_scenarios.py` and `demo_sherlock_investigation.py` use `Path(__file__).resolve().parents[3]` to find the project root. This assumes both files are exactly 3 levels deep, which is fragile. A shared utility or `pyproject.toml`-based package installation would be more robust.

8. **`SCENARIOS` list is defined at module level with no validation.** If someone adds a scenario missing a required key, it will fail at runtime. Consider adding a validation function or using a dataclass.

### Security

9. **Good privacy practices.** The opaque ID approach (`doc_A`, `doc_B`, `doc_C`) is consistently applied. No plaintext content from the encrypted dataset appears in committed files. The privacy assertion tests are a good practice.

10. **Fallback text in `_load_dataset_text` is hardcoded and harmless.** It does not contain any real document content.

### Suggestions (Non-blocking)

- Use a deterministic hash for document-to-index mapping
- Add logging when falling back to sample text
- Validate scenario definitions at import time
- Consider adding accented variants to the privacy check patterns

---

## Cross-PR Observations

### Architecture

The three PRs form a coherent feature stack: orchestrator (#383) -> hypothesis tracker (#384) -> scenarios (#385). However, they have a significant integration gap: `HypothesisTracker` (PR #384) is never actually used by `SherlockModernOrchestrator` (PR #383). The orchestrator calls `invoke_callables._invoke_atms` via `_invoke_safe`, which presumably wraps some other ATMS integration. The new `HypothesisTracker` class is orphaned -- it exists in `agents/core/oracle/` but no code in the orchestrator or investigation modules imports it. This should either be wired in or documented as a standalone utility.

### Test Pattern Duplication

PRs #383 and #385 both define `_run` helpers using the deprecated `asyncio.get_event_loop().run_until_complete()`. They also both import `InvestigationResult`, `OpenDomainResult`, `Attribution`, and `UnifiedAnalysisState`. The mock patch string is duplicated across all three test files. A shared test fixture module would reduce this duplication.

### Deprecation Warning

All three PRs use `from typing import List, Dict, Set, Optional` (PEP 585 deprecated style). Since this is a research project that likely supports Python 3.9+, this is acceptable, but should be flagged for eventual migration to `list[str]`, `dict[str, Any]`, etc.

PR #386, opened 2026-05-01
feat(atms): hypothesis branching with evidence-driven retraction (#359)

jsboigeEpita:
Re-targets PR #384 to main after rebase. Closes #359.

## Summary
- HypothesisTracker with ATMS branching (≥3 hypotheses)
- Evidence-driven retraction on contradiction
- 17 unit tests

## Test plan
- [x] CI passes
- [x] Tests for hypothesis branching
- [x] Original PR #384 was closed when its base branch (feat/357-modern-sherlock-orchestrator) was merged + deleted; this is the rebased version targeting main.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #389, opened 2026-05-01
fix(post-merge-review): address NanoClaw REQUEST CHANGES (#377-#386)

jsboigeEpita:
## Summary

Surgical corrections for review issues that NanoClaw flagged as **REQUEST CHANGES** on PRs #377, #379, #382, #383, #386 — but were merged before the verdicts were addressed. The reviews documented bugs of correctness (not just style); this PR closes them.

### P0 — Correctness (data wrong)

- **#386 `hypothesis_tracker`** — Assumptions can now be shared across hypotheses (was silently overwriting via `Dict[str, str]`); evidence with `supports=` is now a derived ATMS node, not a free-standing assumption (avoids circular ATMS structure); retraction reason cites the **first** contradicting evidence per assumption, not the latest call.
- **#379 `narrative_synthesis`** — Paragraph splitting now uses a ceil-split that preserves all content (was silently dropping the third section when not evenly divisible).
- **#382 `sherlock_modern_orchestrator`** — State is now persisted via `CAPABILITY_STATE_WRITERS` after each phase (the orchestrator was building results but never populating `UnifiedAnalysisState`, defeating the spectacular-analysis promise); `investigate()` resets state for idempotent re-runs; `_solution` declared explicitly in `__init__`.
- **#382 `workflows`** — None-guard around `build_sherlock_modern_workflow()` so a builder failure doesn't insert a broken entry into `WORKFLOW_CATALOG`.
- **#383 `open_domain_investigation`** — Skip the `0.0/10` reliability score line when no arguments were evaluated (was polluting attribution output).
- **#383 `demo_sherlock_investigation`** — Exit code now requires `>=1` claim (`claims_analyzed >= 0` was always True).

### P1 — Silent error swallowing

- **#377 `invoke_callables`** — Replaced bare `except: pass` with `logger.warning(...)` so paragraph-count and ATMS-step failures surface in logs.
- **#376 `invoke_callables`** — Removed `[:60]` truncation + fuzzy substring matching for `fallacy_targets`; uses prefix-based dict lookup instead (avoided false positives across unrelated argument IDs).
- **`sherlock_modern_orchestrator`** — `_invoke_safe` exception logging promoted from DEBUG to WARNING; `_build_result` snapshot failures logged not swallowed.

### Hygiene

- **#375 `fol_logic_agent`** — Added missing `Set` import; `constants`/`variables` typed `Set[str]`; FOL regex filters empty-arg matches and skips arity-0.
- **#377 `test_spectacular_notebook`** — Executed notebook now writes to pytest `tmp_path` (was polluting source tree; broke parallel runs).

## Tests

- 7 new unit tests for `HypothesisTracker` lock in #386 corrections.
- All affected suites green:
  - `test_narrative_synthesis` — 15 passed
  - `test_sherlock_modern_orchestrator` — 20 passed
  - `test_open_domain_investigation` — 18 passed
  - `test_hypothesis_tracker` (new) — 7 passed

## Test plan

- [ ] CI lint + tests pass
- [ ] Sherlock Modern smoke run: `WORKFLOW_CATALOG["sherlock_modern"]` registered, 7 phases
- [ ] HypothesisTracker: 3 hypotheses with shared assumption all retracted on contradiction
- [ ] HypothesisTracker: retraction reason cites first contradicting evidence

### DIFF
--- a/argumentation_analysis/agents/core/logic/fol_logic_agent.py
+++ b/argumentation_analysis/agents/core/logic/fol_logic_agent.py
@@ -19,7 +19,7 @@ Fonctionnalités :
 import logging
 import asyncio
 import inspect
-from typing import Dict, List, Any, Optional, Union, Tuple
+from typing import Dict, List, Any, Optional, Union, Tuple, Set
 from dataclasses import dataclass, field
 from pydantic import PrivateAttr
 
@@ -530,16 +530,22 @@ RÉPONDS EN FORMAT JSON :
         import re
 
         predicates: Dict[str, int] = {}  # name -> arity
-        constants: set = set()
-        variables: set = set()
+        constants: Set[str] = set()
+        variables: Set[str] = set()
 
+        # Note: this regex assumes flat predicate applications with simple
+        # argument lists. Nested predicates like P(Q(x), y) and operators
+        # inside argument lists are not supported (review #375).
         for formula in formulas:
-            # Extract predicate applications: Name(arg1, arg2, ...)
             for match in re.finditer(r"([A-Z][A-Za-z_]*)\(([^)]+)\)", formula):
                 pred_name = match.group(1)
                 args_str = match.group(2)
                 args = [a.strip() for a in args_str.split(",")]
+                # Filter empty args (malformed input like P(, x)).
+                args = [a for a in args if a]
                 arity = len(args)
+                if not arity:
+                    continue
                 if pred_name not in predicates or predicates[pred_name] < arity:
                     predicates[pred_name] = arity
                 for arg in args:
--- a/argumentation_analysis/agents/core/oracle/hypothesis_tracker.py
+++ b/argumentation_analysis/agents/core/oracle/hypothesis_tracker.py
@@ -14,7 +14,7 @@ Key API:
 
 import logging
 from dataclasses import dataclass, field
-from typing import Any, Dict, List, Optional, Set, FrozenSet
+from typing import Any, Dict, List, Optional, Set
 
 from argumentation_analysis.services.jtms.atms_core import ATMS, ATMSNode
 
@@ -56,8 +56,13 @@ class HypothesisTracker:
         self._atms = ATMS()
         self._hypotheses: Dict[str, Hypothesis] = {}
         self._evidence_log: List[Dict[str, Any]] = []
-        self._assumption_to_hypothesis: Dict[str, str] = {}
-        self._contradicted_assumptions: Set[str] = set()
+        # Map assumption → list of hypotheses that include it (review #386:
+        # multiple hypotheses may share an assumption; original code silently
+        # overwrote on collision).
+        self._assumption_to_hypotheses: Dict[str, List[str]] = {}
+        # Map contradicted assumption → first evidence that contradicted it
+        # (review #386: avoid blaming the latest evidence in retraction reason).
+        self._contradicted_by: Dict[str, str] = {}
 
     def create_hypothesis(
         self,
@@ -71,7 +76,7 @@ class HypothesisTracker:
         atms_assumptions = []
         for a in assumptions:
             self._atms.add_assumption(a)
-            self._assumption_to_hypothesis[a] = hyp_id
+            self._assumption_to_hypotheses.setdefault(a, []).append(hyp_id)
             atms_assumptions.append(a)
 
         hyp = Hypothesis(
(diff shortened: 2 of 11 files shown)


clusterManager-Myia:
**Post-hoc review — NanoClaw** APPROVE

Post-merge fixes addressing REQUEST CHANGES from reviews #377-#386. Key corrections: shared assumptions no longer overwrote, retraction reasons blame first contradicting evidence, fallacy matching fixed false positives, narrative ceil-split, box drawing dead code removed, demo exit code fixed. Tests added.

jsboigeEpita:
## Post-hoc Review — Epita #389: fix(post-merge-review) address NanoClaw REQUEST CHANGES (#377-#386)

257+/62-, 11 files. Merged without pre-merge review.

---

### Overview

This PR addresses review findings from NanoClaw's post-hoc reviews of PRs #377-#386. Each fix references the specific review that raised the issue. 11 files touched across 7 distinct fixes + 1 new test file (85 lines).

### Fix-by-fix analysis

#### 1. `fol_logic_agent.py` — Type annotations + malformed input guard (review #375/#386)
- `set` → `Set[str]` for `constants` and `variables`
- Filters empty args from `args_str.split(",")` — handles `P(, x)`
- Skips predicates with 0 arity after filtering
- Adds comment documenting the flat-predicate regex limitation

🟢 **Correct fix.** Empty arg filtering is a real robustness improvement — `P(,x)` would produce `["", "x"]` with arity 2, polluting the constants set with empty strings.

#### 2. `hypothesis_tracker.py` — Shared assumptions + retraction blame (review #386)
- `_assumption_to_hypothesis` (Dict[str, str]) → `_assumption_to_hypotheses` (Dict[str, List[str]])
- `_contradicted_assumptions` (Set) → `_contradicted_by` (Dict[str, str])
- Evidence with supports → derived node (not assumption), avoiding circular ATMS
- Retraction reason now blames the *first* contradicting evidence, not the latest call

🟢 **Critical correctness fix.** The original code silently overwrote `assumption → hypothesis` mappings when multiple hypotheses shared an assumption. The new `_assumption_to_hypotheses.setdefault(a, []).append(hyp_id)` correctly tracks the many-to-many relationship. The `_contradicted_by` dict with `setdefault` ensures the first evidence is recorded, making retraction reasons accurate.

🟢 **Derived vs assumption distinction is important.** The original code always registered evidence as an assumption, then added a justification — creating a contradictory node state. The fix correctly uses `add_node(is_assumption=False)` for derived evidence.

#### 3. `invoke_callables.py` — Fallacy target matching + quality score lookup (review #376)
- `str(target)[:60]` removed; replaced with bidirectional prefix matching (`a.startswith(t) or t.startswith(a)`)
- Quality score lookup: `arg_name[:30] in str(arg_id)` replaced with direct `per_arg_scores.get(arg_name)` + exact fallback
- State writer failures now log warnings instead of silently passing

🟢 **The `arg_name[:30]` substring matching was a genuine bug.** As noted in the review, `arg_name[:30]` would falsely match "counter_argument" against "target_argument" due to shared prefixes. The direct lookup + exact fallback is correct.

🟡 **Bidirectional prefix matching for fallacy targets** (`a.startswith(t) or t.startswith(a)`) is more robust than truncation but could still produce false positives. For example, argument "freedom" would match fallacy target "freedom_of_speech". An exact match or set-based exclusion would be stricter. Acceptable for the current usage where targets are argument IDs, not substrings.

#### 4. `open_domain_investigation.py` — Suppress meaningless score (review #383)
- Skips `Reliability score: 0.0/10 (0 arguments)` when no arguments were evaluated

🟢 **Correct UX fix.** A 0.0/10 score with 0 arguments is noise in the attribution output.

#### 5. `sherlock_modern_orchestrator.py` — State population + idempotency (review #382/#383)
- New `_PHASE_TO_CAPABILITY` dict maps ctx-keys to capability names
- New `_persist_to_state()` calls `CAPABILITY_STATE_WRITERS` after each phase
- New `_reset()` method clears accumulators on `investigate()` entry
- `_invoke_safe` fallback logging upgraded from DEBUG → WARNING
- `_build_result` snapshot failure now logs with WARNING instead of silent `except: pass`
- `_solution` initialized in `__init__` instead of `getattr` fallback

🟢 **This is the most impactful fix in the PR.** The state population gap was the root cause of the "0 populated fields in state_snapshot" issue. Adding `_persist_to_state()` calls after each phase ensures the spectacular workflow's 28+ field coverage promise is met when Sherlock is used as a sub-workflow.

🟢 **Idempotency via `_reset()`** prevents stale state from leaking between sequential investigations — a real bug when `investigate()` was called multiple times on the same orchestrator instance.

#### 6. `workflows.py` — Sherlock None-guard (review #382)
- `build_sherlock_modern_workflow()` return value checked for `None` before catalog insertion

🟢 **Defensive but correct.** If the builder returns `None` (e.g., import failure), the catalog now logs a warning instead of crashing.

#### 7. `narrative_synthesis_plugin.py` — Paragraph split fix (review #377)
- `parts[:len(parts)//2 + 1]` → `parts[:(len(parts)+1)//2]` (ceil split)
- Removes the `+1` offset in the second slice that could drop the middle element

🟢 **Correct fix.** Original code: for 4 parts, `parts[:3]` and `parts[4:]` — drops index 3 (0-indexed). New code: `parts[:2]` and `parts[2:]` — no content loss.

#### 8. Demo scripts (reviews #377/#383)
- Box drawing: dead-code branch removed (`╗` was always unreachable due to precedence)
- Demo exit code: `claims_analyzed >= 0` → `>= 1` (was always true)

🟢 **Both fixes address real bugs.** The `>= 0` check on a `len()` return was always true. The box-drawing precedence bug made the `╗` branch unreachable.

#### 9. New test file: `test_hypothesis_tracker.py` (85 lines)
- 7 tests covering: multiple hypotheses, evidence contradiction, retraction blame accuracy, shared assumptions, derived vs assumption evidence, investigation state snapshot

🟢 **Excellent test coverage for the fixes.** The `test_retraction_reason_blames_first_contradicting_evidence` test directly validates the `_contradicted_by.setdefault` behavior. `test_shared_assumption_across_hypotheses` validates the many-to-many mapping fix.

### Summary

| Fix | Review ref | Severity | Quality |
|-----|-----------|----------|---------|
| FOL type annotations + malformed input | #375/#386 | Medium | Good |
| Hypothesis tracker shared assumptions | #386 | **High** | Excellent |
| Fallacy target matching | #376 | **High** | Good |
| Quality score lookup | #376 | **High** | Good |
| Meaningless score suppression | #383 | Low | Good |
| Sherlock state population | #382 | **Critical** | Excellent |
| Sherlock idempotency | #383 | Medium | Good |
| Sherlock None-guard | #382 | Low | Good |
| Narrative paragraph split | #377 | Medium | Good |
| Demo script fixes | #377/#383 | Low | Good |

🟢 **Overall: well-executed review response.** Each fix is traceable to a specific review finding. The hypothesis tracker and Sherlock state population fixes address genuine correctness issues. Test coverage for the hypothesis tracker is thorough.

**Verdict: APPROVE** — high-quality review response addressing all raised issues with appropriate fixes and tests.

PR #587, opened 2026-05-16
fix(state): add arg_id to PL/FOL/NL-to-logic for reliable formal verification tracking (#571)

jsboigeEpita:
## Summary
- Add optional `arg_id` parameter to `add_propositional_analysis_result()`, `add_fol_analysis_result()`, and `add_nl_to_logic_translation()` for direct ID-based linking
- Update `get_enrichment_summary()` to check PL, FOL, and NL-to-logic translations for formal verification (was only checking NL-to-logic)
- Update `get_argument_profile()` formal results to match by ID first, falling back to text-matching

## Root Cause
`with_formal_verification` was computed only from `nl_to_logic_translations` using text-matching, ignoring `propositional_analysis_results` and `fol_analysis_results` entirely. Even for NL-to-logic, the text-matching was fragile.

## Test Plan
- [x] 8 new tests in `test_formal_verification_tracking_571.py`
- [x] 64 existing shared_state/cross-ref tests GREEN (zero regression)

Closes #571

### DIFF
--- a/argumentation_analysis/core/shared_state.py
+++ b/argumentation_analysis/core/shared_state.py
@@ -703,6 +703,7 @@ class UnifiedAnalysisState(RhetoricalAnalysisState):
         consistent: bool,
         inferences: List[str],
         confidence: float = 0.0,
+        arg_id: Optional[str] = None,
     ) -> str:
         """Add a first-order logic analysis result."""
         fol_id = self._generate_id("fol", self.fol_analysis_results)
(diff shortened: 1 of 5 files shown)


clusterManager-Myia (approved):
Solid. Adding `arg_id` to `add_fol_analysis_result()`, `add_propositional_analysis_result()`, and `add_nl_to_logic_translation()` enables reliable formal verification tracking. ID-first matching with text fallback — backward compatible.

PR #650, opened 2026-05-20
fix(orchestration): parent-harness fallacies register into UnifiedAnalysisState (#648)

jsboigeEpita:
## Summary
Track HH (#648). Root cause is **not** a report-reader key mismatch (the body's original hypothesis) — code-verified:

- The deterministic parent-harness fallback **is already wired** ([conversational_orchestrator.py:621](argumentation_analysis/orchestration/conversational_orchestrator.py#L621)) and **does detect** fallacies on dense texts (>5000 chars).
- But its registration was guarded by `hasattr(state, "add_identified_fallacy")` ([line 1332, pre-fix](argumentation_analysis/orchestration/conversational_orchestrator.py#L1332)). **Runtime-verified** that `UnifiedAnalysisState` has **no** `add_identified_fallacy` (singular) — only `add_fallacy(fallacy_type, justification, target_arg_id)` and `add_identified_fallacies` (plural). The singular method lives only on `StateManagerPlugin` / `PhaseScopedState`.
- So the guard was **always False** here → every detected fallacy was silently dropped on the real state.

### Knock-on (matches corpus C report §3)
Empty `identified_fallacies` → empty Section 3 in the DeepSynthesis report **and** a starved convergence layer (a fallacy signal is 1 of the 5 `compute_argument_convergence` inputs) → corpus C scored 0 named fallacies and only 2 convergence verdicts vs 5 on A/B. One plumbing bug suppressed **both** the surface fallacy metric and the substance convergence metric.

## Fix
- Register via `state.add_fallacy(...)` (present on the real state); keep the singular method only as a fallback for state types that provide it.
- Robust key mapping: `fallacy_type`/`type`/`nom`, `justification`/`explanation`, `source_arg_id`/`target_argument_id`.

## Why the bug slipped through
The existing `test_harness_registers_fallacies` used a bare `MagicMock()` state — where **every** attribute (incl. the nonexistent `add_identified_fallacy`) auto-exists — so it passed while the real state dropped everything. Rewrote it to drive a **real `UnifiedAnalysisState`** and assert fallacies land in `state.identified_fallacies`; added a singular-method-fallback test.

## Test plan
- [x] `pytest tests/unit/argumentation_analysis/orchestration/test_informal_taxonomy_injection_600.py -q` → 23 passed
- [x] The rewritten test fails on the old code (registration skipped → 0 registered), passes on the fix
- [x] `flake8` clean on both files
- [ ] Re-run `scripts/scda_deepsynthesis_vs_baseline.py --corpus C` to confirm Section 3 + convergence recover (follow-up benchmark)

Note: black version drift — local/CI envs differ on pre-existing assert/list formatting; this PR touches **only** new lines, all clean under the CI black, and leaves pre-existing lines (already CI-accepted on main) untouched.

### DIFF
--- a/argumentation_analysis/orchestration/conversational_orchestrator.py
+++ b/argumentation_analysis/orchestration/conversational_orchestrator.py
@@ -1324,25 +1324,43 @@ async def _run_parent_harness_fallback(
             logger.info("Parent harness: no additional fallacies found")
             return None
 
-        # Register any new fallacies into state
+        # Register any new fallacies into state (#648 Track HH).
+        # UnifiedAnalysisState exposes add_fallacy(fallacy_type, justification,
+        # target_arg_id) — NOT add_identified_fallacy (that singular method lives
+        # only on StateManagerPlugin / PhaseScopedState). The previous guard
+        # `hasattr(state, "add_identified_fallacy")` was always False here, so the
+        # harness silently dropped every fallacy it found (empty Section 3 +
+        # starved convergence on dense corpora). Prefer add_fallacy; keep the
+        # singular path only as a fallback for state types that provide it.
         added = 0
         for f in fallacies:
             if not isinstance(f, dict):
                 continue
-            if hasattr(state, "add_identified_fallacy"):
-                try:
-                    state.add_identified_fallacy(
-                        fallacy_type=f.get("fallacy_type") or f.get("nom", "unknown"),
-                        justification=f.get(
-                            "justification",
-                            f"Detected by parent harness (confidence: {f.get('confidence', 'N/A')})",
-                        ),
-                        confidence=f.get("confidence", 0.6),
-                        source_arg_id=f.get("source_arg_id"),
+            fallacy_type = (
+                f.get("fallacy_type") or f.get("type") or f.get("nom") or "unknown"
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (approved):
**APPROVED** — Cluster review (Hermes/po-2026)

Excellent root-cause analysis and fix. Key observations:

✅ **Correct diagnosis**: `hasattr(state, "add_identified_fallacy")` was always False on `UnifiedAnalysisState` — the singular method only exists on `StateManagerPlugin`/`PhaseScopedState`. Classic MagicMock false-positive in tests.

✅ **Proper fix**: Uses `add_fallacy()` (the real API) as primary path, keeps singular as fallback for backward compat. Robust key mapping with 3 aliases per field.

✅ **Test quality**: Rewrote test to use real `UnifiedAnalysisState` instead of MagicMock. Added regression test for fallback path with `spec=["add_identified_fallacy"]`. Asserts on actual state fields (`state.identified_fallacies`).

✅ **No new dependencies, no API changes, backward compatible.**

⚠️ Minor: bare `except` on registration is acceptable here (non-critical, logged) but could narrow to `Exception` for consistency.

PR #661, opened 2026-05-20
feat(logic): instrument PL/FOL formula pipelines with recall metrics (#658)

jsboigeEpita:
## Summary
- Add `pl_metrics` dict to `_invoke_propositional_logic` return values tracking formulas at each pipeline stage (pass1 atoms, pass2 candidates, pre/post sanitize, post-tweety, wide-net extras)
- Add `fol_metrics` dict to `_invoke_fol_reasoning` return values tracking sorts, predicates, candidates, tweety validation, isolation survivors
- Increase per-argument formula generation limit from 6 to 10 for both PL and FOL pipelines
- Add PL whole-text formula generation pass (wide-net) to capture structural patterns missed by per-argument analysis
- 10 unit tests covering metrics presence, counter keys, and args limit verification

## Test plan
- [x] `pytest tests/unit/argumentation_analysis/test_track_mm_formal_recall.py -v` — 10 passed
- [x] Black formatting applied
- [x] Privacy grep clean (no opaque IDs or source references in diff)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #662, opened 2026-05-20
fix(convergence): wire JTMS signal in compute_argument_convergence — Track LL (#656)

jsboigeEpita:
## Summary

- **Root cause**: `_invoke_jtms` named arg beliefs as raw text excerpts (80-char slices), so the `arg_id in name` substring check in `compute_argument_convergence` (signal 4) never matched — JTMS signal was silently dropped for **every corpus**
- **Fix**: prefix each arg belief key with `arg_{i+1}:` in `_invoke_jtms`; update convergence check to `name.startswith(f"{arg_id}:")` to avoid false positives from defeat beliefs (`defeat:fallacy→arg_1:...`) and ATMS entries
- **Tests**: updated 2 existing tests in `test_invoke_jtms.py`; added 11 new tests in `test_track_ll_convergence_depth.py` covering signal firing, false-positive guards, and convergence depth ≥3

## Technical detail

```
Before: arg_beliefs = [text[:80], ...]  # raw text
        check: arg_id in name            # "arg_1" never in "The policy increases..."

After:  arg_beliefs = [f"arg_{i+1}:{t[:66]}", ...]  # prefixed
        check: name.startswith(f"{arg_id}:")          # exact match
```

Defeat beliefs and ATMS entries do not start with `arg_N:` — no false positives.

## Test plan

- [x] `pytest tests/unit/argumentation_analysis/test_track_ll_convergence_depth.py` — 11 tests pass
- [x] `pytest tests/unit/argumentation_analysis/orchestration/test_invoke_jtms.py` — 8 tests pass
- [x] `black --check` — clean
- [x] `flake8` — clean
- [x] Privacy grep — clean

Closes #656

### DIFF
--- a/argumentation_analysis/orchestration/invoke_callables.py
+++ b/argumentation_analysis/orchestration/invoke_callables.py
@@ -970,12 +970,18 @@ async def _invoke_jtms(input_text: str, context: Dict[str, Any]) -> Dict[str, An
             :80
         ]
 
-    arg_beliefs = [_text(a) for a in raw_args[:10]]
+    _raw_arg_texts = [_text(a) for a in raw_args[:10]]
     claim_beliefs = [_text(c) for c in raw_claims[:6]]
 
-    if not arg_beliefs and not claim_beliefs:
+    if not _raw_arg_texts and not claim_beliefs:
         sentences = [s.strip() for s in input_text.split(".") if len(s.strip()) > 10]
-        arg_beliefs = [s[:80] for s in sentences[:8]]
+        _raw_arg_texts = [s[:73] for s in sentences[:8]]
+
+    # Prefix each belief name with its arg_id so compute_argument_convergence
+    # can index JTMS signals by arg_id (startswith "arg_N:").  Without the
+    # prefix, names are raw text excerpts and the arg_id substring check never
+    # matches, silently dropping the JTMS convergence signal for every corpus.
+    arg_beliefs = [f"arg_{i+1}:{t[:66]}" for i, t in enumerate(_raw_arg_texts)]
 
     # ── Step 1: Add argument and claim beliefs (with ExtendedBelief metadata) ─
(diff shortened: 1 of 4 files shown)


clusterManager-Myia (approved):
## Review: PR #662 — fix(convergence): wire JTMS signal in compute_argument_convergence

**Verdict: Approved**

### Core Fix — Correct and Well-Reasoned

The root cause is clearly identified and the fix is appropriate:

- **Before:** `arg_id in belief_name` with raw text excerpts as belief names — the substring check for strings like `arg_1` almost never matched argument text, silently dropping JTMS convergence signals for every corpus.
- **After:** Beliefs are named `arg_N:<text_excerpt[:66]>` and `compute_argument_convergence` uses `name.startswith(f"{arg_id}:")` — precise match, no false positives.

### Strengths

1. **False-positive prevention is thorough.** The new tests cover defeat beliefs (`defeat:...`), cross-arg text references (`arg_2:...arg_1...`), and ATMS entries — all correctly excluded.
2. **Claim beliefs are correctly left unprefixed** — only argument beliefs get the prefix.
3. **Sentence fallback path also gets the prefix** — no edge case missed.
4. **Convergence depth tests** (≥3, ≥4) demonstrate the practical impact of the fix end-to-end.
5. **The colon separator** (`arg_1:`) prevents ambiguity between arg_1 and arg_10 — good design choice.

### Minor Notes

- The truncation change from 80→73 chars for sentence fallback (to fit the prefix within the original budget) is sensible but slightly reduces the text excerpt length. Not a concern in practice.
- The formatting/PEP-8 changes are mixed with the functional change, making the diff larger than necessary, but they're harmless.

### Merge Conflict Warning

⚠️ This PR and #661 both modify the same lines in `invoke_callables.py` and `narrative_synthesis_plugin.py` with identical formatting changes. Merge conflicts are expected — rebase one onto the other after the first lands.

PR #674, opened 2026-05-22
docs(soutenance): update slides to Sprint 12 — Track UU (#673)

jsboigeEpita:
## Summary

- Update soutenance deck from Sprint 4-5 level to Sprint 12 capabilities
- Pipeline diagram: 17 -> 20 phases with new layout (4 rows)
- Replace anecdotal DeepSynthesis comparison with quantitative 3-corpus benchmark data (OO report)
- Add 6 new slides: Convergence 5 signals, Adjudication grounded/claimed, Insight convergent, Benchmarks Sprint 12, Honest Limitations, DeepSynthesis quantitative
- Update existing slides: Progression (Sprint 4->12), Performance, Tests & Qualite, Conclusion, Architecture Lego
- Remove old duplicate Limitations section

## DoD Checklist (#673)

- [x] `slides.md` a jour, coherent avec main `606fa622`
- [x] Aucune regression de contenu
- [x] IDs opaques uniquement (privacy grep CLEAN)
- [x] 20 phases reflected throughout
- [x] 5 convergence signals documented with bug history
- [x] Adjudication grounded/claimed explained
- [x] Benchmark quantitative data from OO report

## Files changed

| File | Change |
|------|--------|
| `docs/soutenance/slides.md` | Updated 10 existing slides, added 6 new slides, removed 1 duplicate |

 Generated with [Claude Code](https://claude.com/claude-code)

PR #675, opened 2026-05-22
feat(llm): reversible OpenRouter provider toggle to bypass OpenAI quota

jsboigeEpita:
## Summary

Adds a **reversible** provider toggle to `create_llm_service`: when both `OPENROUTER_BASE_URL` and `OPENROUTER_API_KEY` are set, chat completions route through OpenRouter (OpenAI-compatible API). Removing `OPENROUTER_BASE_URL` reverts to OpenAI with zero code change.

**Why:** OpenAI account hit a 429 quota exhaustion, blocking #658 live DoD measurement (Volets 1 & 2) and any other live-API work across the cluster. The user confirmed OpenRouter still has quota. Root cause of the prior inability to switch: the `AsyncOpenAI` `client_kwargs` never passed `base_url`, so the SDK always hit OpenAI's official endpoint regardless of env.

## Changes

- **`argumentation_analysis/core/llm_service.py`**: provider-selection block (`use_openrouter = bool(base_url and api_key)`), `base_url` wired into `client_kwargs`, OpenRouter-prefixed model slug via `OPENROUTER_CHAT_MODEL_ID`, and a clearer missing-key `ValueError` that names both providers.
- **`.env.example`**: documents the toggle under the OpenRouter section (commented `OPENROUTER_BASE_URL` + `OPENROUTER_CHAT_MODEL_ID`).
- **`tests/unit/argumentation_analysis/test_llm_service.py`**: `test_create_llm_service_missing_api_key` now clears the OpenRouter env vars too, so it genuinely exercises the "no provider configured" path; assertion updated to the new error message.

## Added scope (post-review, commits 2 & 3)

The first review (`feat(llm)` factory, 3 files) only covered `create_llm_service`. But the toggle alone is insufficient: several call sites built LLM clients **directly from `OPENAI_API_KEY`**, bypassing the factory entirely — so they kept hitting OpenAI's exhausted quota even with OpenRouter configured. These were found and fixed:

- **`argumentation_analysis/orchestration/conversational_orchestrator.py`**: kernel LLM setup now routes through `create_llm_service(service_id="conversational_llm", force_authentic=True)` instead of constructing `OpenAIChatCompletion` from raw env.
- **`argumentation_analysis/orchestration/invoke_callables.py`** — eliminates the raw-SDK bypass at 5 sites:
  - `_get_openai_client()` rewritten to mirror the toggle (covers all raw-`AsyncOpenAI` callers).
  - `_invoke_hierarchical_fallacy` + `_invoke_hierarchical_fallacy_per_argument` (wide-net Phase 1) now build their kernel via `create_llm_service(force_authentic=True)`; the per-arg path shares one service instance.
  - The two inline PL/FOL 2-pass sites (#547 atom inventory, #544 FOL signature) now use `_get_openai_client()`.
  - Left intentionally untouched: `_invoke_self_hosted_llm_fallacy` (self-hosted endpoint path, separate provider).

## Validation

- `pytest tests/unit/argumentation_analysis/test_llm_service.py tests/unit/argumentation_analysis/core/test_llm_service_extended.py -q` → **15 passed**
- `black --check` + `flake8` on all modified source files → clean
- **Live end-to-end:** the #658 consolidated measurement (corpus C, dense ~46K source) ran the full pipeline through OpenRouter for ~50 min with **0 quota / 429 errors** (all LLM calls HTTP 200 to `openrouter.ai`) — confirming every bypass site is now routed. RR convergence depth ≥3 reproduced. (Volet-1 formula survival is a pre-existing Tweety gap, unrelated to this PR; tracked separately.)

## Rebase

Rebased onto `main` @ 861992b2 (now includes #674 + #676). Both files touched by both PRs (`invoke_callables.py`, `conversational_orchestrator.py`) are **hunk-disjoint** with #676's canonical-arg_id signal fix — verified before rebase, clean replay.

## Backward compatibility

Fully backward compatible. With no OpenRouter env vars set, behavior is identical to before (OpenAI official endpoint). The toggle is opt-in and one-line reversible.

Unblocks live-API work cluster-wide (#658 measurement, etc.).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -16,8 +16,13 @@ OPENAI_API_KEY="sk-..."
 OPENAI_CHAT_MODEL_ID="gpt-5-mini"
 # BASE_URL="" # Override OpenAI base URL if needed
 
-# Pour OpenRouter (alternative provider)
+# Pour OpenRouter (alternative provider — bascule de secours si quota OpenAI épuisé)
+# Définir les 2 vars ci-dessous active automatiquement le routage via OpenRouter
+# (compatible OpenAI). Retirer OPENROUTER_BASE_URL pour revenir à OpenAI.
+# Le slug modèle doit être préfixé par le provider (ex. "openai/gpt-5-mini").
 # OPENROUTER_API_KEY="sk-or-v1-..."
+# OPENROUTER_BASE_URL="https://openrouter.ai/api/v1"
+# OPENROUTER_CHAT_MODEL_ID="openai/gpt-5-mini"
 
 # Pour Azure OpenAI
 # AZURE_OPENAI_ENDPOINT="https://your-azure-endpoint.openai.azure.com/"
--- a/argumentation_analysis/core/llm_service.py
+++ b/argumentation_analysis/core/llm_service.py
@@ -108,13 +108,27 @@ def create_llm_service(
             f"Veuillez mettre à jour votre fichier .env avec OPENAI_CHAT_MODEL_ID={new_model_id}"
         )
         model_id = new_model_id
-    # Récupération directe depuis l'environnement
-    api_key = os.environ.get("OPENAI_API_KEY")
+    # --- Sélection du provider : bascule OpenRouter si configurée ---
+    # Si OPENROUTER_BASE_URL est défini (+ OPENROUTER_API_KEY), les appels sont
+    # routés vers OpenRouter (API compatible OpenAI). Sinon, le comportement
+    # OpenAI officiel reste inchangé. Toggle réversible : retirer
+    # OPENROUTER_BASE_URL du .env pour revenir à OpenAI.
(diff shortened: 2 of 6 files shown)


jsboigeEpita (commented):
## APPROVED — OpenRouter provider toggle

3 files, +40/-12. Reversible bypass for OpenAI quota. Tests updated. Same-owner → COMMENT.

jsboigeEpita (commented):
## APPROVED — OpenRouter toggle (HEAD update)

New commits: orchestrator/invoke_callables fixes, expanded tests (+26/-9), slides update merged in.

Same-owner → COMMENT.

clusterManager-Myia (approved):
LGTM. Le toggle OpenRouter est bien concu : reversible, centralise via create_llm_service + _get_openai_client, tests a jour, aucun secret en dur. Tous les chemins LLM passent par la bascule.

PR #702, opened 2026-05-24
fix(counter-args): remove ≤5 cap + sweep all targets — Track GG (#696)

jsboigeEpita:
## Track GG — Counter-argument volume defect (#696)

Parent Epic: #695. Self-picked (workers silent ~3h; po-2025 reassigned to KK #700 to avoid collision).

### The defect (grounded, live-verified)
The pipeline **lost to the zero-shot baseline** on counter-argument volume:
- Sequential (DAG spectacular), corpus C: **5** counter-args vs zero-shot **18** → LOSS
- Conversational, corpus C: **1** counter-arg vs **18** → LOSS

Two root causes in `invoke_callables._invoke_counter_argument`:
- line ~496 `for f in fallacies[:3]` — capped fallacy targets at 3
- line ~513 `if text and len(targets) < 5` — capped total targets at 5

→ at most 5 counter-arguments regardless of how many arguments/fallacies the corpus actually had. The conversational path produced even fewer because the agent runs out of turn budget.

### The fix
1. **DAG path** — removed both caps. Every fallacy AND every argument is now a target. New `_generate_counters_for_targets` **batches** the LLM calls (batch_size=12) so a single mega-prompt no longer silently drops items past the first handful, and one failed batch no longer sinks the rest.
2. **Conversational path** — added a deterministic **5b-6 post-processor** mirroring the existing Dung/modal/ASPIC ones: after the dialogue, sweep every `identified_argument` and ensure ≥1 counter-argument each via `_generate_counter_arguments_from_state`. Gated by `if spectacular and n_args and n_cas < n_args`.

Anti-pendulum: the caps were **deleted**, not swapped for a higher cap.

### Tests (no real API — mocked LLM)
13 new unit tests in `test_counter_argument_caps_gg.py`:
- `_parse_counter_array` tolerates fenced / bare / single-object / garbage JSON
- `_generate_counters_for_targets` sweeps all 25 targets (no cap), splits into 3 batches, isolates batch failures
- `_invoke_counter_argument` builds 16 targets from 6 fallacies + 10 args (was capped at 5)
- `_generate_counter_arguments_from_state` writes ≥1 CA per argument, handles no-args / no-client / plain-string args
- conversational orchestrator stays importable with 5b-6 wired

`black` + `flake8` clean.

### Live verification
`scripts/measure_both_paths_vs_zeroshot.py` (privacy-clean: cardinalities only, gitignored output) is the head-to-head harness. Live A/B/C run on the funded key (ai-01) to follow as a PR comment — DoD: counter-args ≥ zero-shot on every corpus, both paths.

### Files removed and why
None — additive change only.

### DIFF
--- a/argumentation_analysis/orchestration/conversational_orchestrator.py
+++ b/argumentation_analysis/orchestration/conversational_orchestrator.py
@@ -871,6 +871,32 @@ async def run_conversational_analysis(
         except Exception as e:
             logger.warning(f"Belief revision post-processing failed: {e}")
 
+    # 5b-6. Counter-argument enrichment (GG #696)
+    # The conversational dialogue under-produces counter-arguments (often 1),
+    # losing to the zero-shot baseline on the quantitative axis. Sweep every
+    # identified argument and ensure >=1 counter-argument per argument, so the
+    # collaborative path matches the sequential path's coverage.
+    n_args = len(getattr(state, "identified_arguments", []) or [])
+    n_cas = len(getattr(state, "counter_arguments", []) or [])
+    if spectacular and n_args and n_cas < n_args:
+        try:
+            from argumentation_analysis.orchestration.invoke_callables import (
+                _generate_counter_arguments_from_state,
+            )
+
+            ca_result = await _generate_counter_arguments_from_state(state)
+            if ca_result and ca_result.get("added"):
+                conversation_log.append(
+                    {
+                        "phase": "post_processing",
+                        "type": "counter_argument_enrichment",
+                        "added": ca_result["added"],
+                        "targets": ca_result["targets"],
+                    }
+                )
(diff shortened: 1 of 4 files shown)


jsboigeEpita (commented):
**[NanoClaw]** LGTM. Clean fix for the counter-argument volume defect (#696). Key changes: (1) extracted _parse_counter_array, _generate_counters_for_targets (batched, no cap), _generate_counter_arguments_from_state as proper functions; (2) removed the old top-3 fallacies + 5-total caps; (3) new conversational post-processor (5b-6) ensures >=1 counter per argument; (4) comprehensive test coverage (201 lines) with mocked LLM. The benchmark script is a nice addition for measuring pipeline vs zero-shot.

PR #733, opened 2026-05-29
fix(orchestration): hard per-call LLM timeout guard (#730-bis defense-in-depth)

jsboigeEpita:
## Context — live-observed hang

While running the Epic #695 live re-verify on corpus_C, the **conversational-spectacular path hung ~50 minutes** on a single chat-completion call (process: 98s CPU over 56 min wall-clock = pure network-I/O wait, never returning).

Root cause: `_guarded_chat_completion` — the funnel **every** LLM call passes through — issues `client.chat.completions.create(**kwargs)` with **no per-call timeout**. One stalled round-trip hangs the entire analysis run indefinitely.

This is exactly the **defense-in-depth recommended in the #729 root-cause analysis (R248)** but never implemented: po-2023's #729 fix delivered batching (10→4) + `asyncio.gather` parallelization, which reduces *how many* sequential calls happen, but does nothing if a *single* call stalls.

## Fix (anti-pendulum)

A per-call wall-clock bound — **not** a phase-budget raise:

```python
_LLM_CALL_TIMEOUT_S = float(os.environ.get("LLM_CALL_TIMEOUT_S", "300"))
...
if _LLM_CALL_TIMEOUT_S > 0:
    return await asyncio.wait_for(
        client.chat.completions.create(**kwargs), timeout=_LLM_CALL_TIMEOUT_S
    )
```

- **300s default** — far above a legitimate reasoning-model call (<2 min), well below a pathological hang.
- Env-overridable via `LLM_CALL_TIMEOUT_S`; `0` disables.
- On timeout, `asyncio.TimeoutError` propagates and is absorbed by callers' **existing** `except` blocks (gather batch dropped / GG-bis coverage retry) — graceful degradation instead of an infinite hang.

## Verification

- `TestLLMCallTimeoutGuard`: stalled call times out (<0.05s with patched bound), fast call returns normally, disabled (`=0`) does not wrap.
- 18/18 budget-guard tests + 71 orchestration tests (counter-args, 2-pass PL/FOL, budget) green.
- mypy strict (8-file CI scope): clean.

## Files removed and why

None — single source edit + test append (+51 lines).

### DIFF
--- a/argumentation_analysis/orchestration/invoke_callables.py
+++ b/argumentation_analysis/orchestration/invoke_callables.py
@@ -285,6 +285,14 @@ def llm_budget_scope(ceiling: Optional[int] = None) -> Iterator["_LLMBudget"]:
         _llm_budget.reset(token)
 
 
+# Defense-in-depth (#730-bis): hard per-call wall-clock cap so a single stalled
+# network round-trip cannot hang an entire analysis run. Observed live: the
+# conversational-spectacular path hung ~50 min on one unbounded call. Generous
+# default (300s) — far above a legitimate reasoning-model call (<2 min) but well
+# below a pathological hang. Set LLM_CALL_TIMEOUT_S=0 to disable.
+_LLM_CALL_TIMEOUT_S = float(os.environ.get("LLM_CALL_TIMEOUT_S", "300"))
+
+
 async def _guarded_chat_completion(client: Any, **kwargs: Any) -> Any:
     """Single funnel for every LLM chat-completion call (#708 runaway guard).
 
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
[Hermes] — COMMENT_WITH_CONCERNS

Defense-in-depth propre pour #730-bis. Code clean : asyncio.wait_for avec timeout configurable (300s default), 3 tests unitaires couvrant stalled/normal/disabled. Security scan clean.

Concerns :
1. **CI pending** — automated-tests et lint-and-format en cours. Ne pas merger avant résultat CI.
2. asyncio.TimeoutError propagé « handled by callers' existing except blocks » — vérifier que tous les callers de _guarded_chat_completion attrapent bien TimeoutError (ou Exception). Un caller qui ne catcherait que RuntimeError par exemple laisserait le crash remonter.

PR #868, opened 2026-06-01
fix(a-14): guard JTMS router import for graceful degradation (#857)

jsboigeEpita:
## Summary

Fixes the SPOF identified in the A-14 audit: the hard import of `jtms_endpoints` at `api/main.py:11` caused the entire FastAPI app to crash if the module failed to import (missing JPype/JVM, DLL crash, etc.).

## Changes

**`api/main.py`**:
- Wrapped `jtms_endpoints` import in `try/except ImportError`
- Added `_JTMS_AVAILABLE` boolean flag
- Guarded `initialize_jtms_services()` call in `startup_event` with try/except
- Conditionally mount `jtms_router` only when `_JTMS_AVAILABLE is True`
- All failures log warnings instead of crashing

**`tests/unit/api/test_jtms_import_guard.py`** (11 new tests):
- 7 source structure tests (static analysis verifies guard is syntactically present)
- 4 runtime behavior tests (constructed FastAPI apps with/without JTMS)
- Existing 12 JTMS mount tests still pass (23/23 total)

## DoD (#857)
- [x] JTMS import wrapped in try/except with warning log
- [x] Router mount conditional on availability flag
- [x] startup_event guards init call
- [x] Test demonstrates app boots without JTMS routes
- [x] Anti-pendule: JTMS mount is KEPT, just resilient (not removed)

## À valider par lutilisateur

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -8,8 +8,27 @@ from .proposal_endpoints import proposal_router
 from .mobile_endpoints import mobile_router
 from .websocket_routes import ws_router
 from .agent_routes import agent_router
-from argumentation_analysis.api.jtms_endpoints import jtms_router, initialize_jtms_services
 from argumentation_analysis.core.bootstrap import initialize_project_environment
+
+# JTMS endpoints — optional, graceful degradation if import fails (#857).
+# The JTMS module depends on JPype/JVM; if unavailable, the API boots
+# without JTMS routes rather than crashing entirely.
+try:
+    from argumentation_analysis.api.jtms_endpoints import (
+        jtms_router,
+        initialize_jtms_services,
+    )
+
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (approved):
[Hermes] — APPROVED

- Security: clean
- JTMS import guard is well-structured: try/except ImportError with fallback to None, conditional router mount, guarded startup init with catch-all Exception handler.
- Tests cover both static source analysis (7 tests) and runtime behavior (4 tests) without importing api.main.
- CI: all green.
- Cause racine (#857): JPype/JVM dependency makes JTMS import fail on environments without JVM. Fix is complete with graceful degradation.

PR #891, opened 2026-06-02
feat(evaluation): orchestration mode comparison harness - 8 modes, 3 corpora

jsboigeEpita:
Implements orchestration mode comparison harness dispatched in R313/R314.

## Summary
New script `scripts/compare_orchestration_modes.py` benchmarks 8 orchestration modes on 3 benchmark corpora. Markdown + JSON output.

## Usage
```bash
python scripts/compare_orchestration_modes.py --modes pipeline_light conversation_deterministic --output report.md
```

## Collision-guard: Does NOT touch --mode, orchestrators, or service_manager.

### DIFF
--- a/scripts/compare_orchestration_modes.py
+++ b/scripts/compare_orchestration_modes.py
@@ -1,0 +1,493 @@
+"""Compare orchestration modes on the same corpus.
+
+Runs multiple orchestration modes (pipeline, conversational, hierarchical,
+cluedo_baseline, cluedo_extended, conversation_deterministic) on benchmark
+texts and produces a markdown comparison report with aggregate metrics.
+
+Usage:
+    # Compare all available modes on benchmark texts
+    python scripts/compare_orchestration_modes.py
+
+    # Specific modes only
+    python scripts/compare_orchestration_modes.py --modes pipeline conversational
+
+    # Save report to file
+    python scripts/compare_orchestration_modes.py --output report.md
+
+    # Dry run (show which modes would run, no execution)
+    python scripts/compare_orchestration_modes.py --dry-run
+
+Modes are skipped gracefully if their orchestrator is unavailable (e.g. missing
+LLM, JVM not initialized, or mode not yet wired).
+"""
+
+import argparse
+import asyncio
+import json
+import logging
+import sys
+import time
+from dataclasses import asdict, dataclass, field
+from datetime import datetime
+from pathlib import Path
+from typing import Any, Dict, List, Optional
+
+# Add project root to path
+PROJECT_ROOT = Path(__file__).parent.parent
+if str(PROJECT_ROOT) not in sys.path:
+    sys.path.insert(0, str(PROJECT_ROOT))
+
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s [%(levelname)s] [%(name)s] %(message)s",
+    datefmt="%H:%M:%S",
+)
+logger = logging.getLogger("orchestration_mode_harness")
+
+# Reduce noisy loggers
+for _name in ("httpx", "openai", "semantic_kernel", "urllib3"):
+    logging.getLogger(_name).setLevel(logging.WARNING)
+
+
+# ── Benchmark texts (opaque IDs, no raw content in reports) ──────────────
+
+BENCHMARK_TEXTS = {
+    "corpus_A": (
+        "Le Premier ministre a déclaré que la réforme des retraites est nécessaire "
+        "car tous les pays européens l'ont déjà faite. C'est un argument d'autorité "
+        "qui ne tient pas compte des différences structurelles entre les systèmes. "
+        "De plus, affirmer que « si nous n'agissons pas maintenant, le système "
+        "s'effondrera dans cinq ans » est un appel à la peur classique. "
+        "Les syndicats rétorquent que le gouvernement utilise un sophisme naturaliste "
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

## Review: feat(evaluation): orchestration mode comparison harness (PR #891)

Single-file harness (`scripts/compare_orchestration_modes.py`, 493 lines) with 8 orchestration mode runners, a benchmark corpus of 3 French argumentation texts, markdown/JSON report generation, and a basic pytest suite (4 tests).

---

### Concerns

#### 1. Sequential execution defeats purpose of comparison (logic)

`run_all()` iterates modes and corpora in a nested loop with `await runner(text, corpus_id)`. LLM-backed modes (`conversational`, `pipeline`) could take 30-120s each. With 8 modes x 3 corpora = 24 runs, wall-clock time could exceed 15 minutes with no parallelism. The modes are independent — there is no shared mutable state between runs. Consider `asyncio.gather` or `asyncio.TaskGroup` to run modes in parallel per corpus:

```python
# Parallel per corpus, sequential across corpora (to limit memory):
for corpus_id in corpora:
    tasks = [runner(text, corpus_id) for mode, runner in runners.items()]
    corpus_results = await asyncio.gather(*tasks, return_exceptions=True)
```

#### 2. `state_fill_rate` is inconsistent across modes (logic correctness)

The core comparison metric (`state_fill_rate`) is computed differently depending on the mode:
- **pipeline/conversational**: `non_empty_fields / total_fields` — a structural fill ratio (0.0-1.0 range)
- **conversation_deterministic**: `conv_state.get("state", {}).get("score", 0)` — a raw score, semantics unknown

These are not comparable metrics, yet the report presents them side-by-side in the "Cross-Mode Comparison" section under "**Highest state fill**". The reader will draw incorrect conclusions. Either normalize to a common scale, or label them distinctly in the report.

#### 3. `phases_completed` and `phases_total` are hardcoded in `conversation_deterministic_mode` (logic correctness)

Line 218-219:
```python
phases_completed=3,  # informal + fol + synthesis
phases_total=3,
```

This is a hardcoded assumption about the internal pipeline structure. If `ConversationOrchestrator.run_orchestration()` changes its phase structure, this silently reports wrong data. The actual orchestrator's result should drive these values, even if accessed differently than the pipeline mode.

#### 4. No timeout on individual mode runs (robustness)

`run_all()` catches broad exceptions in the outer `try/except` (line 437-443), but there is no timeout on individual `await runner(...)` calls. If an LLM-backed mode hangs (API timeout, slow response, retry loop), the entire harness blocks indefinitely. Consider `asyncio.wait_for(runner(text, corpus_id), timeout=120)`.

#### 5. `sys.path` manipulation is fragile (code quality)

Line 52-53:
```python
if str(PROJECT_ROOT) not in sys.path:
    sys.path.insert(0, str(PROJECT_ROOT))
```

This is standard for standalone scripts, but the subsequent `from compare_orchestration_modes import ...` in tests (line 24) relies on the test's `sys.path` already being correct. The test does `from compare_orchestration_modes import BENCHMARK_TEXTS` without the `sys.path` setup the script itself does. This works only because pytest's rootdir discovery happens to include the scripts directory — fragile and non-obvious.

#### 6. Error truncation loses diagnostic context (robustness)

Line 264: `error=str(e)[:200]` and line 442: `error=f"Exception: {str(e)[:200]}"`. Truncating to 200 chars can cut off the actual useful part of the traceback. Consider capturing `repr(e)` or `traceback.format_exc()` with a slightly higher limit (500 chars), or at least including the exception type in the truncated string.

#### 7. `conversational_mode` has no error handling (robustness)

`run_conversational_mode` (line 159-192) and `run_pipeline_mode` (line 123-156) have zero try/except — any exception propagates to the caller. Meanwhile `run_hierarchical_mode` (line 231-264) properly wraps its execution. Either all runners should handle their own errors, or all should propagate (and the caller handles uniformly). The current inconsistency means `conversational` failures produce different-shaped error data than `hierarchical` failures.

#### 8. Test coverage gaps

- No test for `run_all()` end-to-end (even with mocked runners)
- No test for the report markdown format (only checks substring presence, not table structure)
- No test for `--dry-run` CLI path
- No test for unknown mode/corpus handling
- `test_conversation_deterministic_produces_result` actually calls the real `ConversationOrchestrator` — this is an integration test, not a unit test, and will fail without the JVM or required dependencies

#### 9. `BENCHMARK_TEXTS` corpus IDs are opaque but content is hardcoded French (minor)

The docstring says "opaque IDs, no raw content in reports" but the actual content is hardcoded French political/argumentation texts in the script. If the goal is to avoid leaking real document content, this is fine. But these texts are too similar in structure (all French political argumentation with fallacies) to serve as a broad benchmark. Consider adding at least one corpus with different domain/structure to avoid overfitting the comparison to a single rhetorical pattern.

---

### No security issues

No secrets, credentials, or unsafe patterns detected. No external API calls are made directly (they go through the orchestrator modules). No file writes outside of user-specified `--output` path.

---

### Verdict: COMMENT_WITH_CONCERNS

The harness is structurally sound and well-documented, but the inconsistent `state_fill_rate` metric (concern #2) undermines the core purpose of the comparison. The hardcoded phases (#3) and lack of timeouts (#4) are correctness/robustness issues that should be addressed before this is used for benchmarking decisions. The sequential execution (#1) is a practical concern but not blocking.

PR #894, opened 2026-06-02
docs(design): parametric integration specs -- tracks 1 (fallacy-tier) & 3 (shield-preset)

jsboigeEpita:
## Design Specs — Parametric Integration Tracks 1 & 3

Two design specifications for the parametric integration north-star (R311). These are **design documents only** — no code changes. Each spec includes current state analysis, proposed CLI/pipeline/API integration, implementation plan, testing plan, and risks.

### Track 1: `--fallacy-tier` Selector (`docs/architecture/design_specs/TRACK-01_fallacy_tier_selector.md`)

**Current state**: 6 distinct fallacy detection strategies exist, but no centralized selector. Pipeline always runs full LLM-based detection.

**Proposal**: 4 tiers mapping to detector combinations:
| Tier | Detectors | LLM? | Speed |
|------|-----------|------|-------|
| `taxonomy` | TaxonomySophismDetector only | No | ~ms |
| `hybrid` | FrenchFallacyAdapter (symbolic+NLI) | Optional | ~1s |
| `llm` (default) | FallacyWorkflowPlugin + per-arg | Yes | ~10-30s |
| `full` | All strategies merged | Yes (max) | ~30-60s |

**Effort**: S (~210 LOC). Default = `llm` (no breaking change).

### Track 3: `--shield-preset` Selector (`docs/architecture/design_specs/TRACK-03_shield_preset_selector.md`)

**Current state**: 4 presets exist (`basic`/`advanced`/`output_only`/`strict`), registry + invoke callable + REST endpoint all exist. Only gap: CLI argument and workflow phase.

**Proposal**: Add `--shield-preset {off,basic,advanced,output_only,strict}` to CLI. Default = `off` (no change).

**Effort**: S (~68 LOC). Trivial wiring.

### À valider par l'utilisateur

1. **Fallacy tier default**: `llm` (current behavior) or `hybrid` (cheaper)?
2. **Shield default**: `off` (no overhead) or `basic` (zero-cost heuristic guard)?
3. **Tier naming**: `taxonomy/hybrid/llm/full` or `light/standard/deep/exhaustive`?

### Validation

- [x] No code changes (design docs only)
- [x] Privacy: opaque IDs only
- [x] Cleanup Gate: no deletions
- [x] Research grounded in actual codebase (file:line references verified)

### DIFF
--- a/docs/architecture/design_specs/TRACK-01_fallacy_tier_selector.md
+++ b/docs/architecture/design_specs/TRACK-01_fallacy_tier_selector.md
@@ -1,0 +1,215 @@
+# Design Spec — Track 1: Fallacy-Tier Selector (`--fallacy-tier`)
+
+**Date**: 2026-06-02
+**Author**: Claude Code @ myia-po-2023
+**Base**: `841ed091` (main, post-hierarchical reactivation)
+**Effort**: S (1 session)
+**Parent**: `docs/architecture/parametric_integration_map.md` (PR #885)
+
+---
+
+## Verdict
+
+**GO** — The codebase already has 6 distinct fallacy detection strategies and partial tier toggling (`FrenchFallacyAdapter` constructor args). The gap is a **single CLI/pipe selector** that maps a user-facing tier name to the correct combination of detectors and pipeline phases. Effort is small because the detectors exist; we only need the routing layer.
+
+---
+
+## 1. Current State
+
+### 1.1 Six Detection Strategies (existing)
+
+| # | Strategy | Type | Location | LLM? |
+|---|----------|------|----------|------|
+| 1 | FrenchFallacyAdapter (5-tier hybrid) | Neural + Symbolic + LLM | `adapters/french_fallacy_adapter.py` | Optional |
+| 2 | FallacyWorkflowPlugin (hierarchical taxonomy) | LLM iterative deepening | `plugins/fallacy_workflow_plugin.py` | Yes |
+| 3 | Per-argument parallel detection | LLM parallel | `invoke_callables.py:_invoke_hierarchical_fallacy_per_argument` | Yes |
+| 4 | TaxonomySophismDetector | Lexical keyword match | `agents/core/informal/taxonomy_sophism_detector.py` | No |
+| 5 | ContextualFallacyDetector | Rule-based contextual | `agents/tools/analysis/new/contextual_fallacy_detector.py` | No |
+| 6 | ComplexFallacyAnalyzer | Structural pattern | `agents/tools/analysis/complex_fallacy_analyzer.py` | No |
+
+### 1.2 Current Selection Mechanism
+
+**No centralized selector.** Each detector is instantiated independently:
+- Pipeline always runs `_invoke_hierarchical_fallacy` (full LLM-based, strategy 2+3)
+- `FrenchFallacyAdapter` has per-tier toggles (`enable_symbolic`, `enable_nli`, etc.) but **is not wired into pipeline phases**
+- `AnalysisDepth` enum (`BASIC`/`STANDARD`/`COMPREHENSIVE`/`EXPERT`) exists in `fallacy_family_analyzer.py` but only governs the family analyzer, not other detectors
+
+### 1.3 Gaps
+
+1. No pipeline-level parameter to choose between lightweight vs deep detection
+2. `FallacyWorkflowPlugin` constants (`MAX_DEPTH_PER_BRANCH=8`, `MAX_BRANCHES=4`) are hardcoded
+3. `AnalysisDepth` enum not propagated to pipeline
+4. `FrenchFallacyAdapter` tier toggling is isolated — not wired into pipeline phases
+
+---
+
+## 2. Proposed Design
+
+### 2.1 Tier Definitions
+
+| Tier | Label | Detectors | LLM Calls | Speed | Use Case |
+|------|-------|-----------|-----------|-------|----------|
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Design-only PR — two specs for parametric CLI selectors. Thorough and well-referenced overall. Several issues below.

---

### Track 1: Fallacy-Tier Selector

**Strengths:**
- File/line references verified against actual codebase: `_invoke_hierarchical_fallacy` at line 3680, `_invoke_ai_shield` at 6118, `ai_shield_service` at 667, `FrenchFallacyAdapter` constructor args (enable_symbolic/enable_nli/enable_llm) confirmed at lines 1222-1225. All 6 detectors exist at the referenced paths.
- `AnalysisDepth` enum claim checked: not found in `invoke_callables.py` — spec says it's in `fallacy_family_analyzer.py`. This path is not referenced in section 1.1 but is implied; minor gap in traceability but not a blocker.
- No-breaking-change strategy is sound: default `llm` matches current pipeline behavior.
- Tier definitions are coherent and map cleanly to existing detector capabilities.

**Concerns:**

1. **API code sample uses deprecated `regex=` parameter.** Track 1 section 2.5:
   ```python
   fallacy_tier: str = Query("llm", regex="^(taxonomy|hybrid|llm|full)$"),
   ```
   In FastAPI/Pydantic v2, `regex=` is deprecated in favor of `pattern=`. Since this is a design spec, it should reference the current API framework version the project uses. Same issue appears in Track 3 section 2.4. Worth noting so implementation doesn't copy a deprecated pattern.

2. **`_invoke_hybrid_fallacy()` spec says `enable_llm=False, enable_nli=True` — but section 2.1 table says hybrid has "0-1 (optional NLI)".** The code sample hardcodes NLI on. The tier definition should clarify: is NLI always enabled for hybrid (requiring ~600MB model download), or should it be truly optional? If NLI is optional, the design needs a sub-toggle or the table should say "0" for the hybrid tier.

3. **`_invoke_full_fallacy()` merge strategy is underspecified.** "Merges by `taxonomy_pk`, keeps highest confidence" — but `TaxonomySophismDetector` returns string labels while `FallacyWorkflowPlugin` returns structured LLM outputs with reasoning chains. The merge logic needs more than a PK match: it needs a canonical label mapping. This is the hardest part of the `full` tier and deserves its own subsection.

4. **Section 1.2 line 3 says "Pipeline always runs `_invoke_hierarchical_fallacy`"** — but `invoke_callables.py` line 3680 shows it's an async function called from the callable registry, not hardcoded into every pipeline. The spec conflates "default workflow calls it" with "always runs." Minor but worth precision for implementers.

---

### Track 3: Shield Preset Selector

**Strengths:**
- Accurate inventory of existing wiring: presets.py verified (80 lines, 4 presets: basic/advanced/output_only/strict), registry entry at 667, invoke callable at 6118, shared_state confirmed.
- Decision to use Option B (context-driven) over Option A (workflow phase) for v1 is pragmatic.
- Default `off` preserves current behavior cleanly.

**Concerns:**

5. **`presets.py` docstring is stale.** The module docstring (lines 2-5) lists only 3 presets: "basic, advanced, output_only" — omitting "strict" which was added later. The Track 3 spec correctly lists all 4, but the upstream source it references has this inconsistency. Should be flagged as a pre-requisite fix (docstring update) before implementation.

6. **`fail_open` logic for `strict` is hardcoded in the CLI spec:**
   ```python
   "fail_open": args.shield_preset != "strict",
   ```
   But `load_preset("strict", fail_open=False)` already hardcodes `fail_open=False` in presets.py (line 68: `fail_open=False`). So the CLI layer sets `fail_open=False` AND the preset also sets it — double enforcement. Not a bug, but the spec should acknowledge that `fail_open` in `load_preset()` for `strict` is immutable, making the CLI-level override misleading.

7. **Line range claim "lines 21-81" for presets.py** — file is 80 lines total, so the accurate range is 1-80 (or 21-80 for the function body). Off by one; cosmetic but reflects on spec precision.

8. **Option A recommends shield as "first phase (before extraction)"** — this contradicts the spec's own section 2.3 which says `fail_open=True` (continue on failure). If shield runs first and fails, the pipeline continues to extraction with unvalidated input. If the intent is to block malicious input, shield should run *after* extraction (filtering LLM output) rather than before (blocking raw input). The `basic` preset with heuristic-only makes more sense as a pre-extraction filter, but `output_only` is explicitly post-LLM. The spec doesn't address this phase-ordering tension.

---

### Cross-track issues

9. **PR title says "tracks 1,2,3" but only Track 1 and Track 3 are included.** Track 2 is missing without explanation. The PR body says "Tracks 1 & 3" which is accurate, but the title is wrong. This should be fixed to avoid confusion.

10. **Both specs import `Query` from FastAPI in code samples but neither specifies the import.** Since these are design docs targeting implementers, `from fastapi import Query` should be explicit in the code blocks (current usage: `Query("llm", regex=...)` without import context).

11. **Both specs reference `api/agents.py` (Track 1) and `api/shield_endpoints.py` (Track 3) as separate files** but the Track 3 code sample adds `shield_preset` to the *same* endpoint as `fallacy_tier` (`/api/v1/agents/full-analysis`). If Track 1 adds it to `api/agents.py` and Track 3 adds it to the same endpoint, the two specs should cross-reference or the endpoint parameter accumulation should be documented in one place.

---

### Verdict: COMMENT_WITH_CONCERNS

Well-researched design specs with accurate codebase references. The main substance issues are: the hybrid tier's NLI ambiguity (#2), the full-tier merge strategy being underspecified (#3), the shield phase-ordering tension (#8), and the PR title mismatch (#9). The rest are precision and documentation issues that should be cleaned up before implementation begins.

PR #895, opened 2026-06-02
feat(adapters): abs_arg_dung provider for dung_extensions capability (#893)

jsboigeEpita:
## Summary

Implements issue #893 — Registers the student Dung library (`abs_arg_dung/`) as a selectable provider in ServiceDiscovery for the `dung_extensions` capability.

## Architecture

```
DungStudentProvider (adapter)
  └─ wraps EnhancedDungAgent (student lib, sanctuary)
      └─ asyncio.to_thread bridge
          └─ 4 semantics: grounded, preferred, stable, complete
```

Registered as `dung_extensions_student` (quality_score=0.6) alongside the native `AFHandler` engine (11 semantics, quality_score=0.8).

## Key features
- **Sanctuary rule**: `abs_arg_dung/` source is NEVER modified
- **Graceful fallback**: returns unavailable dict when JVM not started or library missing
- **Shape normalization**: output matches the `_invoke_dung_extensions` format (extensions, statistics, arguments, attacks)
- **ServiceDiscovery integration**: `register_dung_student_provider(registry)` for CapabilityRegistry

## Files
- **NEW**: `argumentation_analysis/adapters/dung_student_provider.py` (295 LOC)
- **NEW**: `tests/unit/argumentation_analysis/adapters/test_dung_student_provider.py` (7 tests)

## Tests
```
7 passed in 3.93s
```

## Collision-guard
Does NOT touch `--mode`, orchestrators, `service_manager`, or `registry_setup.py`. Purely additive adapter.

## Files removed and why
None.

## Known issues documented (NOT patched)
- `abs_arg_dung/cli.py` bare imports break package-mode execution
- Student lib only supports 4/11 Dung semantics vs native engine

### DIFF
--- a/argumentation_analysis/adapters/dung_student_provider.py
+++ b/argumentation_analysis/adapters/dung_student_provider.py
@@ -1,0 +1,279 @@
+"""Adapter for abs_arg_dung as an alternative Dung extensions provider.
+
+Registers the student Dung library (abs_arg_dung) as a selectable provider
+in ServiceDiscovery for the ``dung_extensions`` capability, alongside the
+native AFHandler engine.  The student library is treated as a **sanctuary**
+(never modified); this adapter wraps its public API from the outside.
+
+Architecture:
+- ``DungStudentProvider`` wraps ``EnhancedDungAgent`` via ``asyncio.to_thread``
+  (the agent is synchronous JPype-based).
+- Shape normalisation: converts the agent's output to the same dict format
+  as the native ``_invoke_dung_extensions`` callable.
+- Graceful fallback if JVM is not started or the student library is missing.
+
+Sanctuary rule: ``abs_arg_dung/`` source is NEVER modified.  All adaptation
+happens in this file.  Known issues (e.g. ``cli.py`` bare imports) are
+documented but NOT patched in-place.
+
+Issue: #893
+"""
+
+import asyncio
+import logging
+from typing import Any, Dict, List, Optional, Tuple
+
+logger = logging.getLogger(__name__)
+
+# Semantics supported by EnhancedDungAgent (subset of the 11 in af_handler)
+SUPPORTED_SEMANTICS = [
+    "grounded",
+    "preferred",
+    "stable",
+    "complete",
+]
+
+
+class DungStudentProvider:
+    """Wraps abs_arg_dung.EnhancedDungAgent as a selectable Dung provider.
+
+    The student library is synchronous and requires a running JVM.
+    This provider wraps calls via ``asyncio.to_thread`` for async compatibility.
+    """
+
+    def __init__(self):
+        self._agent = None
+        self._available: Optional[bool] = None
+
+    @property
+    def provider_name(self) -> str:
+        return "abs_arg_dung_student"
+
+    @property
+    def quality_score(self) -> float:
+        """Quality score for ServiceDiscovery ranking.
+
+        Lower than native AFHandler (0.8) since the student lib only
+        supports 4 semantics vs 11 in the native engine.
+        """
+        return 0.6
+
+    @property
+    def capabilities(self) -> List[str]:
+        return ["dung_extensions"]
+
+    def is_available(self) -> bool:
+        """Check if the student library is importable and JVM is running."""
+        if self._available is not None:
+            return self._available
+
+        try:
+            import jpype
+
+            if not jpype.isJVMStarted():
+                logger.info("DungStudentProvider: JVM not started, unavailable")
+                self._available = False
+                return False
+
+            from abs_arg_dung.enhanced_agent import EnhancedDungAgent
+
+            self._available = True
+            logger.info("DungStudentProvider: available (EnhancedDungAgent importable)")
+            return True
+        except ImportError as e:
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

## Review: PR #895 — feat(adapters): abs_arg_dung provider for dung_extensions capability

### Summary

Adds a `DungStudentProvider` adapter wrapping the student `abs_arg_dung` library as a selectable `dung_extensions` provider in ServiceDiscovery, alongside the native AFHandler. Clean architecture: sanctuary pattern (no student code modified), lazy init, async via `asyncio.to_thread`, graceful degradation.

### Concerns

**1. `_build_framework` does not verify arguments exist in the AF after adding them**

```python
# Lines 119-128
for arg_name in arguments:
    arg = agent.Argument(arg_name)
    agent.af.add(arg)

for attacker, attacked in attacks:
    attacker_arg = agent.Argument(attacker)
    attacked_arg = agent.Argument(attacked)
    attack = agent.Attack(attacker_arg, attacked_arg)
    agent.af.add(attack)
```

`agent.af.add(arg)` may silently no-op on duplicates. More importantly, the attack loop creates *new* `Argument` objects for attacker/attacked without verifying they were successfully added. If an attack references an argument not in `arguments`, a phantom `Argument` gets created in the attack tuple — behavior is then undefined (depends on whether the JPype bridge compares by identity or by name). This should at minimum validate that `attacker` and `attacked` are in the `arguments` set, or catch the resulting Java-side exception.

**2. `DungStudentProvider()` is instantiated fresh in `invoke_dung_student` on every call**

```python
# Lines 241-242
provider = DungStudentProvider()
return await provider.compute_extensions(arguments, attacks)
```

And separately in `register_dung_student_provider`:
```python
# Line 256
provider = DungStudentProvider()
```

These are two separate instances. The registration creates one (for `is_available()` check), then `invoke_dung_student` creates a new one per call — meaning `_available` is re-checked and `_agent` (the `EnhancedDungAgent`) is lazily re-created each time. This is correct for the `_available` cache (class-level would be wrong if JVM state changes), but the per-call `_agent` instantiation is wasteful if the student library has non-trivial init cost. Consider a module-level or registry-stored singleton, or at minimum document the design choice.

**3. No thread-safety on `_agent` / `_available`**

`_compute_extensions_sync` runs in a thread via `asyncio.to_thread`, but `_get_agent()` and `_build_framework()` both read/write `self._agent` without any lock. If `compute_extensions` is called concurrently (e.g., two requests hitting the adapter simultaneously), two threads could race on `self._agent = EnhancedDungAgent()` and on `agent.af`. For a teaching repo this is unlikely to be hit, but it should be noted.

**4. `_compute_extensions_sync` output shape inconsistency for error cases**

When a semantics computation fails, the entry is `{"error": str(e)[:200]}`. When it succeeds, it has `{"extensions", "count", "sizes", "all_members"}`. The `semantics_computed` stat at line 186 counts by checking `"count" in v`, which works but is fragile — any future semantics entry with a "count" key for a different reason would be miscounted. Consider a dedicated `"status": "ok"/"error"` field instead.

**5. `quality_score` lower than native, but no ranking logic is visible in this diff**

The comment says native is 0.8, student is 0.6. The ranking logic lives in ServiceDiscovery which is not in this diff — this is fine, but worth verifying that ServiceDiscovery actually uses `quality_score` from metadata (line 269 stores it) rather than from a provider property.

**6. Test `test_unavailable_when_import_fails` patches `sys.modules` but doesn't clear the cached `_available` properly**

```python
# Lines 335-339
with patch.dict("sys.modules", {"abs_arg_dung.enhanced_agent": None}):
    provider._available = None
    result = provider.is_available()
    assert result is False
```

The `jpype.isJVMStarted()` mock returns `True`, then `from abs_arg_dung.enhanced_agent import EnhancedDungAgent` is attempted. The `patch.dict` sets the module to `None`, so `is_available()` should return `False` from the `ImportError`. This works, but `patch.dict` context managers can have surprising interaction with import caching in Python 3.9+. The test is functional but brittle — worth a comment noting the dependency on import system internals.

**7. `attacks` type hint is `List[Tuple[str, str]]` but the output dict uses `list(attacks)`**

If `attacks` is a list of tuples, `list(attacks)` preserves the tuples. The output `attacks` field will contain tuples, not lists. Not a bug per se, but inconsistent with JSON-serializable expectations if downstream consumers expect lists of lists.

### Minor notes

- The docstring says "Issue: #893" — good traceability.
- The known-issues block at the bottom of the file (lines 278-286) is a good practice for documenting without patching sanctuary code.
- Tests cover the main paths (unavailable, metadata, registration skip) but do not test the happy path of actual extension computation. This is understandable since it requires a live JVM + student lib, but a mock-based test of `_compute_extensions_sync` returning shaped data would improve coverage.

### Verdict

**COMMENT_WITH_CONCERNS**

The adapter is well-structured and follows good patterns (sanctuary, lazy init, graceful degradation). The concerns above are real but none are blocking for a teaching repo — items 1 (phantom arguments in attacks) and 2 (per-call instantiation) are the most worth addressing before merge. The rest are defensive improvements.

PR #919, opened 2026-06-03
feat(api): orchestration_mode selector with real dispatch (#917)

jsboigeEpita:
## Summary

Expose `orchestration_mode` on `POST /api/workflow/custom` with **real dispatch** to the correct orchestrator — not just context propagation.

| Mode | Entry function | Dispatch line |
|------|---------------|---------------|
| pipeline (default) | `run_unified_analysis()` | proposal_endpoints.py dispatch |
| conversational | `run_conversational_analysis()` | imported from conversational_orchestrator |
| hierarchical | `run_hierarchical_analysis()` | imported from hierarchical/orchestrator |
| sherlock_modern | `SherlockModernOrchestrator.investigate()` | imported from sherlock_modern_orchestrator |

## Anti-inerte R321

This is NOT just adding a field and propagating to context. The endpoint **dispatches** to the real orchestrator function based on mode:
- `conversational` calls `run_conversational_analysis` — NOT `run_unified_analysis`
- `hierarchical` calls `run_hierarchical_analysis` — NOT `run_unified_analysis`
- `sherlock_modern` calls `SherlockModernOrchestrator.investigate` — NOT `run_unified_analysis`

Each dispatch test mocks BOTH the target orchestrator AND `run_unified_analysis`, then asserts:
1. Target orchestrator WAS called
2. `run_unified_analysis` was NOT called (no silent fallback)

## Files changed

- `api/proposal_models.py` — `orchestration_mode` field (Literal, default=pipeline)
- `api/proposal_endpoints.py` — dispatch logic in `run_custom_workflow`
- `tests/unit/api/test_parametric_api.py` — 27 tests (was 20), including 4 dispatch-verification tests

## Tests

```
27/27 passed in 11.80s
```

New tests:
- `TestOrchestrationModeModel` — 3 model validation tests (all modes, invalid rejected, default)
- `TestOrchestrationModeDispatch` — 4 endpoint tests:
  - default dispatches to pipeline
  - conversational dispatches to conversational (NOT pipeline)
  - hierarchical dispatches to hierarchical (NOT pipeline)
  - sherlock_modern dispatches to sherlock (NOT pipeline)

## Collision-guard

api/ only, 0 conflict with argparse (po-2025).

## A valider par l utilisateur

RAS - backward-compatible defaults, no breaking change. cluedo excluded (separate system).

Closes #917

### DIFF
--- a/api/proposal_endpoints.py
+++ b/api/proposal_endpoints.py
@@ -198,12 +198,15 @@ async def list_capabilities():
 
 @proposal_router.post("/workflow/custom", response_model=WorkflowResult)
 async def run_custom_workflow(request: CustomWorkflowRequest):
-    """Run a custom analysis workflow on arbitrary text."""
-    try:
-        from argumentation_analysis.orchestration.unified_pipeline import (
-            run_unified_analysis,
-        )
+    """Run a custom analysis workflow on arbitrary text.
 
+    Dispatches to the correct orchestrator based on orchestration_mode:
+    - pipeline (default): run_unified_analysis (sequential DAG)
+    - conversational: run_conversational_analysis (multi-agent dialogue)
+    - hierarchical: run_hierarchical_analysis (strategic planning → Lego)
+    - sherlock_modern: SherlockModernOrchestrator (investigation)
+    """
+    try:
         # Build context from parametric selectors (#903, #910)
         context: Dict[str, Any] = {}
         context["fallacy_tier"] = request.fallacy_tier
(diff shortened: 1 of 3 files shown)


jsboigeEpita (commented):
**[NanoClaw]**

## Code Quality: Good

Clean dispatch pattern with literal string matching against a `Literal` type -- no injection surface. Lazy imports inside each branch keep startup lean and avoid importing unavailable modules. The `else` default to pipeline is correct.

## Concern 1: Parametric context silently dropped for non-pipeline modes (MEDIUM)

Lines 207-215 build a `context` dict from `fallacy_tier`, `shield_preset`, `vote_method`, and `consensus_threshold`, but only the `pipeline` branch passes it to its orchestrator. The `conversational`, `hierarchical`, and `sherlock_modern` branches ignore it entirely.

A caller who sets `orchestration_mode: "conversational"` alongside `fallacy_tier: "deep"` gets no feedback that their selector was ignored. This is a footgun for API consumers.

Suggestion: either (a) propagate relevant context fields to the non-pipeline orchestrators, or (b) return a warning/header when non-default selectors are provided with a non-pipeline mode.

## Concern 2: sherlock_modern result truncation (LOW)

The `InvestigationResult` dataclass has 7 fields (`trace`, `reasoning_chain`, `agents_used`, `agent_count`, `hypotheses`, `solution`, `state_snapshot`). The dispatch code only forwards 4 of them (`trace`, `solution`, `agents_used`, `hypotheses`), silently dropping `reasoning_chain`, `agent_count`, and `state_snapshot`.

This is information loss for the caller. Consider forwarding the full dataclass dict instead of picking fields.

## Concern 3: SherlockModernOrchestrator() without state (LOW)

Line 55 instantiates `SherlockModernOrchestrator()` without a `UnifiedAnalysisState`. This means `self.state is None` inside `investigate()`, so `_persist_to_state` is a no-op and `state_snapshot` in `_build_result()` is always `None`.

This is likely acceptable for a minimal API dispatch, but worth noting -- the orchestrator runs without state tracking, which means no state-based fallbacks or snapshot. If this is intentional, a one-line comment would help.

## Concern 4: conversational return type assumption (NIT)

The `conversational` and `hierarchical` branches assign their return to `result` and rely on the existing `isinstance(result, dict)` check downstream. This works because both functions are documented to return `Dict[str, Any]`, but the mock in the sherlock test uses a `MagicMock` object rather than a dict -- the test works only because the dispatch code constructs a plain dict from `inv_result`. The other three mode tests never exercise the `isinstance` check with a non-dict result. Minor test coverage gap.

## Tests: Good

27/27 tests, 4 dispatch-specific tests each verifying both positive dispatch AND negative pipeline-not-called assertion. The `assert_not_called` guard is exactly the right pattern for preventing silent fallback.

## Security: Clean

No dynamic imports, no eval/exec, no user input in import paths. The `orchestration_mode` is validated by Pydantic `Literal` before reaching the endpoint. No concerns.

## Summary

The dispatch logic is solid, backward-compatible (default=pipeline), and well-tested. The main actionable concern is the silent context drop for non-pipeline modes (Concern 1) -- callers will lose parametric selectors without warning. The sherlock result truncation (Concern 2) and missing state (Concern 3) are lower priority but worth tracking.

PR #926, opened 2026-06-03
feat(parametric): wire Dung provider selection into _invoke_dung_extensions (#908)

jsboigeEpita:
## Summary

Wire the Dung provider selection into  so that `context["dung_provider_hint"] == "abs_arg_dung_student"` delegates to the student Dung provider instead of the native AFHandler.

**North-Star parametric integration (#78), Track 9.**

### `argumentation_analysis/orchestration/invoke_callables.py`
- Added provider delegation block in `_invoke_dung_extensions` (after arg/attack extraction, before AFHandler try block)
- When `dung_provider_hint == "abs_arg_dung_student"`: instantiates `DungStudentProvider`, calls `compute_extensions()`
- Graceful fallback: if provider returns `status=unavailable` or raises, falls back to native AFHandler
- Unknown hint values ignored (backward-compat)
- Fix: em-dash (U+2014) in `_invoke_ai_shield` docstring caused SyntaxError on some encodings

### `tests/unit/argumentation_analysis/test_dung_provider_selector.py` (NEW — 9 tests)
- `TestDungProviderContext` (2): context propagation, zero-pollution
- `TestDungProviderConsumer` (4): no hint = native, student hint = delegation, unavailable = fallback, unknown = ignored
- `TestDungStudentProvider` (3): provider_name, quality_score, capabilities

## Test Results

```
9 passed, 7 warnings in 5.82s
```

Plus verified 23 formal_extension tests still pass (no regression).

## Parametric Integration Queue

| Track | Issue | Status | Sole owner |
|-------|-------|--------|------------|
| 7 | #899 `--fallacy-tier` | MERGED | po-2025 |
| 8 | #900 `--fol-solver` | MERGED | po-2025 |
| 9 | #904 `--counter-strategy` | MERGED | po-2025 |
| 10 | #905 `--formal-extension` | MERGED | po-2025 |
| **11** | **#908 Dung provider wiring** | **THIS PR** | **po-2025** |
| 12 | #914 `--mode cluedo` | NEXT | po-2025 |

## Collision Guard

po-2025 is sole owner of argparse `run_orchestration.py`. This PR only touches `invoke_callables.py` (consumer-side wiring). No overlap with po-2023 API surface work.

### DIFF
--- a/argumentation_analysis/orchestration/invoke_callables.py
+++ b/argumentation_analysis/orchestration/invoke_callables.py
@@ -5467,8 +5467,11 @@ async def _invoke_dung_extensions(
                 DungStudentProvider,
             )
 
-            student_provider = DungStudentProvider()
-            result = await student_provider.compute_extensions(arguments, attacks)
+            student_provider = DungStudentProvider()  # type: ignore[no-untyped-call]
+            # attacks is List[List[str]] from _generate_attacks_from_args;
+            # compute_extensions expects List[Tuple[str, str]] — convert
+            _typed_attacks = [(a[0], a[1]) for a in attacks if len(a) >= 2]
+            result = await student_provider.compute_extensions(arguments, _typed_attacks)
             if result.get("status") == "unavailable":
                 logger.warning(
                     "DungStudentProvider unavailable, falling back to native AFHandler"
--- a/argumentation_analysis/orchestration/invoke_callables.py
+++ b/argumentation_analysis/orchestration/invoke_callables.py
@@ -5448,12 +5448,48 @@ async def _invoke_dung_extensions(
     stable, complete, admissible, conflict_free, semi_stable, stage, cf2,
     ideal, naive.
     Falls back to pure-Python computation when JVM is unavailable.
+
+    #908: If ``context["dung_provider_hint"]`` is set to
+    ``"abs_arg_dung_student"``, delegates to the student Dung provider
+    instead of the native AFHandler.
     """
     # 1. Extract arguments from upstream phases
     arguments = _extract_arguments_from_context(input_text, context)
 
     # 2. Build attack relations from fallacies and counter-arguments
     attacks = _generate_attacks_from_args(arguments, context)
+
+    # #908: Provider selection — delegate to student provider if hinted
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]** Review PR #926

## Summary

Clean provider delegation wiring for Track 11 of the parametric integration (#78). The new block in `_invoke_dung_extensions` correctly intercepts `context["dung_provider_hint"] == "abs_arg_dung_student"` after argument/attack extraction and before the native AFHandler block, with two fallback paths: graceful (provider returns `status=unavailable`) and exception-based (any error from the student provider). Unknown hint values are silently ignored, preserving backward compatibility.

## Concerns

**1. Silent data loss on malformed attacks (low)**
```python
_typed_attacks = [(a[0], a[1]) for a in attacks if len(a) >= 2]
```
Any attack with fewer than 2 elements is silently dropped. This is likely fine given that `_generate_attacks_from_args` should always produce pairs, but adding a `logger.debug` for the dropped count would make debugging easier if upstream behavior changes.

**2. Trivial context-propagation tests (nit)**
`test_default_no_hint` and `test_hint_set_in_context` test raw dict operations (`"x" not in {}`, `context["x"] == "y"`), not the actual wiring. They don't exercise `_invoke_dung_extensions` at all. Consider either removing them or merging into the consumer tests as assertions on the context dict before/after invocation.

**3. Trace entry on short-circuit (nit)**
The trace entry is only added when `result.get("arguments")` is truthy. If the student provider returns a valid response with no arguments (empty list), the trace entry is skipped. Consider using `result.get("arguments") is not None` instead, so an empty-but-valid result still gets traced.

## No issues found

- No security concerns (no secrets, no user input flowing unsanitized)
- Fallback logic is sound — both the `unavailable` status check and the broad `except Exception` are appropriate for an optional delegation path
- The em-dash fix in `_invoke_ai_shield` is a valid SyntaxError prevention on restricted encodings
- No regressions: 9 new tests pass, 23 existing formal_extension tests confirmed green
- Imports are lazy (inside the `if` block), so no startup cost when the student provider is unused

PR #931, opened 2026-06-04
feat(capstone): brick-health harness (#928)

jsboigeEpita:
Part of #923. Tests 13 invoke callables in isolation. 12/13 PASS.


jsboigeEpita (commented):
test review access

PR #933, opened 2026-06-04
feat(parametric): implement --mode cluedo selector (#914)

jsboigeEpita:
## Summary

Implements `--mode cluedo` as an argparse-selectable orchestration mode, dispatching to `run_cluedo_oracle_game()`.

Closes #914

### `argumentation_analysis/run_orchestration.py`
- Added `"cluedo"` to `--mode` argparse choices (line 338)
- Added help text for cluedo mode (line 346)
- Added `elif mode == "cluedo"` dispatch branch (line 677-731):
  - Builds SK `Kernel` with LLM service
  - Derives `initial_question` from `text_content` (first 500 chars) or defaults to investigation prompt
  - Calls `run_cluedo_oracle_game(kernel, initial_question)`
  - Displays investigation/solution/trace summary
  - Saves JSON output if `--output` specified

### `tests/unit/argumentation_analysis/test_cluedo_mode_dispatch.py` (NEW)
- `TestCluedoModeArgparse`: 2 tests — argparse accepts "cluedo", default stays "pipeline"
- `TestCluedoModeDispatch`: 2 tests — dispatch calls `run_cluedo_oracle_game()`, `initial_question` incorporates text

All 4 tests pass.

## Testing

```bash
conda run -n projet-is-roo-new python -m pytest tests/unit/argumentation_analysis/test_cluedo_mode_dispatch.py -v
# 4 passed
```

## Checklist
- [x] mypy strict (no new errors)
- [x] Tests pass (4/4)
- [x] Rebased on `origin/main`

### DIFF
--- a/argumentation_analysis/run_orchestration.py
+++ b/argumentation_analysis/run_orchestration.py
@@ -320,12 +320,14 @@ Exemples:
             "pipeline",
             "conversational",
             "hierarchical",
+            "cluedo",
             "sherlock_modern",
         ],
         default="pipeline",
         help="Mode d'orchestration: pipeline (séquentiel, défaut), "
         "conversational (agents dialoguent via AgentGroupChat), "
         "hierarchical (strategic planning → Lego execution), "
+        "cluedo (investigation Sherlock-Watson-Oracle, #914), "
         "sherlock_modern (investigation multi-agent, #357)",
     )
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

## Review: #933 — feat(parametric): implement --mode cluedo selector (#914)

Overall this is a clean, well-structured addition that follows the existing dispatch patterns in `run_orchestration.py`. A few items worth addressing.

### Security
- No secrets, hardcoded keys, eval/exec, or user input passed to `subprocess`/`os.system`. Clean.

### Logic — Kernel setup concern (medium)

Lines 650-659 of the dispatch branch build a kernel and conditionally attach an LLM service, but the logic has a subtle issue:

```python
if llm_service and hasattr(llm_service, "service_id"):
    from argumentation_analysis.core.llm_service import create_llm_service

    svc = create_llm_service(service_id="cluedo_mode")
```

1. **`llm_service` is imported but the conditional re-creates a brand-new service via `create_llm_service()`** rather than using `llm_service` itself. This means the outer `llm_service` (from `setup_environment()`) is checked but never actually attached to the cluedo kernel — a fresh one is created instead. If `llm_service` is already available from `setup_environment()`, it should probably be reused, or at minimum the condition should be explained as intentional.

2. **`hasattr(svc, "add_to_kernel")` branch is dead code.** The `create_llm_service()` return type is `OpenAIChatCompletion | AzureChatCompletion | ChatCompletionClientBase` — none of these have an `add_to_kernel` method. Semantic Kernel connectors use `kernel.add_service()`, which is the `else` branch. The `add_to_kernel` path will never execute, making the `hasattr` check misleading. Consider removing the dead branch and unconditionally calling `kernel.add_service(svc)`.

3. **No error handling if `create_llm_service()` returns `None`.** The `svc` could be `None` (e.g. missing API keys), and the subsequent `hasattr(svc, ...)` would silently pass without attaching any service. The kernel would then be empty when passed to `run_cluedo_oracle_game()`. An explicit `if svc is None` guard with a log warning would be safer.

### Logic — `run_cluedo_oracle_game` failure (minor)

```python
results = await run_cluedo_oracle_game(kernel=kernel, initial_question=initial_question)
```

There is no `try/except` around this call. If the cluedo orchestrator raises (e.g. agent setup failure, LLM timeout), the entire `main()` crashes without a user-friendly message. Other modes in the file (conversational, hierarchical) also lack this, but it would be a good opportunity to add a targeted catch here since cluedo is a complex multi-agent workflow.

### Logic — `text_content` can be empty string (low)

The text source validation at line ~461 checks `if not text_content` which would catch `None`, but not `""` (it would be falsy though, so it would catch it). However, the cluedo branch has its own check:

```python
if text_content
```

This correctly handles `None` and `""`. However, note that the main function already enforces `--file`, `--text`, or `--ui` and returns early if `text_content` is falsy, so the `else` fallback (default French prompt) in the cluedo branch is effectively unreachable in normal CLI usage. This is fine — defensive coding — but worth noting for clarity.

### Test quality

**Good:**
- Argparse tests correctly validate that "cluedo" is accepted as a choice and that the default remains "pipeline".
- Mock targets match real import paths (`argumentation_analysis.orchestration.cluedo_extended_orchestrator.run_cluedo_oracle_game`).
- Mock target matches real function signature (`kernel: Kernel, initial_question: str`).
- `AsyncMock` is correctly used for the async function.

**Issues:**
- **Test weak assertions:** `test_dispatch_calls_run_cluedo_oracle_game` only checks `assert_called_once()` and then a redundant `call_kwargs is not None or call_count > 0`. The `or` means the second condition is never reached if `call_kwargs` is not None. It should use `and` or separate assertions. More importantly, it doesn't verify the _arguments_ passed (kernel instance, initial_question content).
- **`setup_environment` mock scope:** Patching `argumentation_analysis.run_orchestration.setup_environment` should work, but the mock returns `MagicMock(service_id="test")` — the cluedo branch checks `hasattr(llm_service, "service_id")` which is `True` for any MagicMock, so this correctly exercises the service-creation path. However, `create_llm_service` is not mocked inside the cluedo branch, so the test will actually call `create_llm_service()` during execution. This could cause test failures in CI environments without API keys. Consider patching `create_llm_service` as well, or patching `argumentation_analysis.run_orchestration.create_llm_service` inside the import scope.
- **No test for:**
  - `--output` flag saving to file
  - Empty kernel (service creation failure) path
  - The display/print formatting
  - `SystemExit` handling is silently swallowed — if `main()` exits with code != 0 for a legitimate error, the test passes anyway, masking real failures

### Consistency with existing patterns

The cluedo dispatch follows the established pattern well:
- Import at branch entry (same as sherlock_modern, conversational, hierarchical)
- Display block with `=` separator (same format)
- Output save block with `mkdir(parents=True)` (same pattern)
- Logging in French (consistent)

One minor inconsistency: the `--output` log message is in English (`"Results saved to {output_path}"`) while all other modes use French (`"Résultats sauvegardés dans {output_path}"`). The print block also uses English headers (`"Cluedo Investigation Results"`) while other modes use French. This should match the codebase convention.

### Summary

The dispatch wiring is correct and well-placed. The main actionable items are:
1. Remove the dead `add_to_kernel` branch and add a `None` guard on `svc`.
2. Fix the test to patch `create_llm_service` to avoid real API calls in CI.
3. Normalize output language to French (log message + print headers).
4. Strengthen test assertions to verify call arguments.

PR #955, opened 2026-06-05
fix(fol): eprover→tweety binary-absent fallback (#949)

jsboigeEpita:
## Summary

Closes #949 (Epic #947, Phase 2 — FB-4)

When the configured external FOL solver (eprover/prover9) is absent, `FOLHandler` now gracefully falls back to the in-JVM Tweety FOL reasoner — mirroring the modal pattern (`modal_handler.py:36-48`). A `solver_fallback` flag is returned for degradation tracking.

### `fol_handler.py` (+59/-6)
- `fol_check_consistency()`: try/except around eprover/prover9 → fallback to tweety on `RuntimeError`
- `fol_query()`: same pattern, returns `(entailed, solver_fallback)` tuple
- New `_fol_check_consistency_with_tweety_fallback()` method with `solver_fallback=True` flag
- `logger.warning` on fallback (detectable in dev/CI logs)

### `test_fol_handler.py` (new, 6 tests)
| Test | Result | Gate |
|------|--------|------|
| eprover absent → tweety fallback | PASS | `solver_fallback=True` |
| eprover present → no fallback | PASS | 2-tuple result |
| tweety direct path → no fallback | PASS | 2-tuple result |
| query fallback same pattern | PASS | `solver_fallback=True` |
| query no fallback when eprover works | PASS | `solver_fallback=False` |
| both solvers absent → RuntimeError | PASS | no silent failure |

## Local run
```
6 passed, 5 warnings in 4.11s
```

## DoD checklist
- [x] When eprover binary is absent, FOL uses in-JVM Tweety (verified by test)
- [x] `solver_fallback` flag returned for degradation tracking
- [x] WARNING log on fallback for dev/CI visibility
- [x] No regression on FOL per-formula isolation (`invoke_callables.py` untouched)
- [x] Both solvers absent → RuntimeError (not silent)

## Anti-pendule
Solver *selection* fallback only — does not touch per-formula isolation logic.

## Privacy
Synthetic formulas only, opaque IDs.

### DIFF
--- a/argumentation_analysis/agents/core/logic/fol_handler.py
+++ b/argumentation_analysis/agents/core/logic/fol_handler.py
@@ -257,16 +257,52 @@ class FOLHandler:
     async def fol_check_consistency(self, belief_set):
         """
         Checks if an FOL knowledge base is consistent using the configured solver.
+
+        When the configured external solver (eprover/prover9) is unavailable
+        (binary absent, RuntimeError), falls back to the in-JVM Tweety FOL
+        reasoner — mirroring the modal pattern (modal_handler.py:36-48).
+        Returns a ``solver_fallback`` flag in the result so callers can track
+        degradation.
         """
         self.logger.debug(
             f"Performing FOL consistency check with solver: {settings.solver.value}"
         )
         if settings.solver == SolverChoice.PROVER9:
-            return await self._fol_check_consistency_with_prover9(belief_set)
+            try:
+                return await self._fol_check_consistency_with_prover9(belief_set)
+            except RuntimeError as e:
+                self.logger.warning(
+                    f"Prover9 unavailable ({e}), falling back to Tweety FOL reasoner"
+                )
+                return await self._fol_check_consistency_with_tweety_fallback(belief_set)
         elif settings.solver == SolverChoice.EPROVER:
-            return await self._fol_check_consistency_with_eprover(belief_set)
+            try:
+                return await self._fol_check_consistency_with_eprover(belief_set)
+            except RuntimeError as e:
+                self.logger.warning(
+                    f"EProver unavailable ({e}), falling back to Tweety FOL reasoner"
+                )
+                return await self._fol_check_consistency_with_tweety_fallback(belief_set)
         else:
             return await self._fol_check_consistency_with_tweety(belief_set)
+
+    async def _fol_check_consistency_with_tweety_fallback(self, belief_set):
+        """Attempt Tweety FOL consistency check; returns solver_fallback=True on success."""
+        try:
+            result = await self._fol_check_consistency_with_tweety(belief_set)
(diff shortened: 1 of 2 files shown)


jsboigeEpita (commented):
**[NanoClaw]**

## Verdict: REQUEST CHANGES needed — breaking return-type changes on `fol_query` and `fol_check_consistency`

The fallback logic itself is well-designed: try/except around external solvers, graceful degradation to Tweety, `solver_fallback` flag for observability, WARNING logs, and RuntimeError on total failure. Tests for the new behavior are thorough. Security scan: clean.

**However, the PR introduces silent breakage across existing callers without updating them.**

---

### 1. `fol_query` return type change: `bool` -> `tuple[bool, bool]` (BREAKING)

`query_executor.py:134-138` does:
```python
result = self._tweety_bridge.fol_handler.fol_query(belief_set, query)
if isinstance(result, bool):  # <-- this is now NEVER true
    ...
```
A tuple is never `isinstance(result, bool)`, so **every FOL query routed through `QueryExecutor`** now silently falls through to the error branch and returns `(None, result)`. No exception, no warning — just `None` accepted as the answer.

### 2. `fol_check_consistency` asymmetric return type (latent breakage)

On fallback the method returns a 3-tuple `(is_consistent, msg, solver_fallback)`, but normally returns a 2-tuple `(is_consistent, msg)`. Any caller that unpacks with `is_consistent, msg = handler.fol_check_consistency(...)` will get `ValueError: too many values to unpack` **only when the external solver is absent** — the exact scenario this PR is meant to handle gracefully. This is a code smell; the failure mode is triggered precisely in the fallback path.

### 3. Existing tests will break (not updated in this PR)

- `test_eprover_spass_integration.py:102` asserts `result is True` — now gets a tuple.
- `test_query_executor.py:174` mocks `fol_query.return_value = True` — the mock no longer matches the real return type, masking the breakage in (1).
- `test_fol_handler_config.py:33` integration test for `fol_query_solver_dispatch` — same issue.

### 4. Return-type inconsistency between the two methods

`fol_query` always returns a 2-tuple `(entailed, solver_fallback)`. `fol_check_consistency` returns a 2-tuple normally but a 3-tuple on fallback. These should follow the same pattern — either always include the flag or document why they differ.

---

### Recommended fix

1. **Update all callers** of `fol_query` and `fol_check_consistency` to handle the new return types (or add a compatibility wrapper).
2. **Make return types uniform**: both methods should always return the `solver_fallback` flag, even when it's `False`. This eliminates the asymmetric unpacking hazard.
3. **Update or skip existing tests** that assert on the old return type.
4. Consider a deprecation path if downstream consumers are not all in this repo.

The fallback mechanism itself is sound and the test coverage for the new behavior is good — the issue is that the interface change is not propagated to consumers.

PR #984, opened 2026-06-06
fix(pipeline): add narrative_synthesis phase to standard workflow (#981)

jsboigeEpita:
## Summary

The narrative_synthesis capability was fully registered (registry + invoke callable + state writer) but the default standard workflow never included it as a phase. This caused --mode pipeline (defaulting to --workflow standard) to silently skip narrative synthesis.

Conversational mode worked because the LLM autonomously discovers and calls the NarrativeSynthesisPlugin kernel_function — no pipeline phase needed.

## Root cause

build_standard_workflow() in workflows.py had no narrative_synthesis phase. Both build_full_workflow() and build_spectacular_workflow() included it.

## Fix

Added narrative_synthesis phase to build_standard_workflow() with the same dependency set used in full:
- quality, hierarchical_fallacy, counter, jtms, dung_extensions

All 5 dependencies already exist in the standard workflow. Phase is optional=True matching the existing pattern.

## Verification

Built workflow standard_analysis with 15 phases (was 14). narrative_synthesis phase present with correct deps. No circular dependency, no import errors.

## Files changed

argumentation_analysis/orchestration/workflows.py (+14 lines)

## Related

- Roadmap P0 no.1 (narrative_synthesis pipeline activation)
- Epic 947 Phase 3 coverage

### DIFF
--- a/argumentation_analysis/orchestration/workflows.py
+++ b/argumentation_analysis/orchestration/workflows.py
@@ -237,6 +237,20 @@ def build_standard_workflow() -> WorkflowDefinition:
             depends_on=["counter"],
             optional=True,
         )
+        # Narrative synthesis (#981): produce a structured narrative from all
+        # analysis results. Depends on phases that populate convergence signals.
+        .add_phase(
+            "narrative_synthesis",
+            capability="narrative_synthesis",
+            depends_on=[
+                "quality",
+                "hierarchical_fallacy",
+                "counter",
+                "jtms",
+                "dung_extensions",
+            ],
+            optional=True,
+        )
         # Local LLM offline analysis (#835 A-10) — skipped if endpoint absent
         .add_phase(
             "local_llm",


clusterManager-Myia (commented):
**[NanoClaw]**

LGTM. Clean, minimal, correct fix.

**What this does:** Adds the `narrative_synthesis` phase to `build_standard_workflow()`, which was the only workflow missing it despite the capability being fully registered (registry + invoke callable + state writer).

**Correctness:**
- Dependencies (`quality, hierarchical_fallacy, counter, jtms, dung_extensions`) match `build_full_workflow()` exactly -- consistent with the stated intent.
- `optional=True` matches every other optional phase in the standard workflow.
- All 5 dependency phases already exist in the standard DAG, so no new edges are created that could introduce cycles.
- Phase is inserted before `local_llm` (end of chain), which is appropriate -- narrative synthesis aggregates upstream results, it does not block downstream work.
- DAG topology remains acyclic: extract -> {neural_detect, hierarchical_fallacy, nl_to_logic, quality} -> {pl, fol, dung_extensions, counter} -> {jtms, governance, debate, **narrative_synthesis**} -> local_llm.

**Observation (non-blocking):**
- `build_spectacular_workflow()` uses a slightly different dep set (`quality, jtms, atms, dung_extensions`) -- it includes `atms` instead of `hierarchical_fallacy` and `counter`. The standard workflow does not have an `atms` phase, so the chosen deps are the correct adaptation. The PR description correctly calls this out as using the "same dependency set as full".
- The `local_llm` phase that follows has `depends_on=["extract"]` only -- `narrative_synthesis` and `local_llm` are effectively independent (both terminal/leaf phases), so insertion order does not matter functionally, but placing synthesis before local_llm is the right convention.

No secrets, no behavioral risk. The PR description verification (15 phases, no circular deps) is accurate.

PR #997, opened 2026-06-07
fix(test): restore FOL semantic invariant — Fallacious->consistent=False (#996)

jsboigeEpita:
## Summary
Follow-up to #990 (PR merged as d14476ee). The unconditional type+non-empty assertions from #986 guaranteed non-vacuous execution but lost the semantic check that `Fallacious` predicates must produce `consistent=False` (original #976 invariant).

## Changes
- Restored conditional semantic assertion behind the unconditional guards
- Cleaned up docstring (merged #986 + #996 references)

## DoD (#996)
- [x] Unconditional type+non-empty assertions remain (no regression of #990)
- [x] Semantic invariant (Fallacious->consistent=False) restored
- [x] Docstring cleaned (#986/#996)
- [x] 3/3 FOL value-gate tests PASS

## Anti-pendule
- The conditional is NOT vacuous: the unconditional guards ensure formulas>0
- No return to vacuous-pass (#980 pattern)

**Fichier**: test_value_gates.py only. No invoke_callables.py (reserved po-2025).

### DIFF
--- a/tests/unit/argumentation_analysis/test_value_gates.py
+++ b/tests/unit/argumentation_analysis/test_value_gates.py
@@ -1397,9 +1397,9 @@ class TestFOLValueGate:
         consistent=True when formulas contain 'Fallacious' predicates
         (which indicate undermined arguments -- consistent should be False).
 
-        #986: Assertions are UNCONDITIONAL — the test must fail loudly if
-        the fallback produces formulas without any 'Fallacious' marker (which
-        would make the original conditional assertion silently pass at-vacuous).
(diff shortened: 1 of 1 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Clean test fix. The new conditional assertion is well-structured:

- The unconditional guards from #986 guarantee the test is never vacuous, so the `if any("Fallacious" in f ...)` check cannot silently pass at zero.
- Comment cross-references (#986, #996, #976) are accurate and helpful.
- The assertion message includes the formulas for debuggability.

No issues. LGTM.

PR #998, opened 2026-06-07
feat(benchmark): FR↔EN fallacy alias map for FB-8 yardstick scoring (#995)

jsboigeEpita:
## Summary
The FB-8 benchmark scorer expected EN canonical markers (circular_reasoning, ad_populum, false_dilemma...) but the pipeline detects fallacies with FR labels (Faux dilemme, Raison de la majorité, Appel à l'émotion...). This caused false-negative scoring on D1/D3/D6/D7.

## Changes
- Added `FALLACY_FR_EN_ALIASES` constant (30+ mappings, derived from `taxonomy_full.csv` text_fr→text_en columns)
- Added `_normalize_fallacy_type()` function applied in the scorer's fallacy_type collection
- Both `family` and `fallacy_type` fields are now normalized through the alias map
- EN snake_case markers pass through unchanged

## DoD (#995)
- [x] Alias table covers 8 families + yardstick markers (D1-D10)
- [x] Scorer matches FR-detected fallacies against EN yardstick markers
- [x] Test: 'Faux dilemme' → 'false_dilemma' scores correctly
- [x] Test: D3 (Populist Rhetoric) passes from MISSED→PARTIAL via 'Raison de la majorité'→ad_populum
- [x] 9/9 tests PASS

## Anti-pendule
- No broad substring matching — explicit family→marker aliases only
- Taxonomy CSV checked first: carries both text_fr and text_en columns

**Files**: run_benchmark_fb8.py + new test file. No invoke_callables.py (reserved po-2025).

### DIFF
--- a/scripts/run_benchmark_fb8.py
+++ b/scripts/run_benchmark_fb8.py
@@ -268,6 +268,70 @@ def _extract_full_state(snapshot: Dict[str, Any]) -> Dict[str, Any]:
 
 
 # ---------------------------------------------------------------------------
+# FR↔EN Fallacy alias map (#995)
+# ---------------------------------------------------------------------------
+
+# Derived from taxonomy_full.csv (text_fr → text_en canonical).
+# Covers the 8 families + all yardstick markers (D1-D10).
+# Each key is a normalised (lowercased) FR label produced by the pipeline;
+# each value is the EN canonical marker expected by the yardstick scorer.
+FALLACY_FR_EN_ALIASES: Dict[str, str] = {
+    # Insufficiency family
+    "généralisation hâtive": "hasty_generalization",
+    "rationalisation": "rationalization",
+    "appel à l'ignorance": "appeal_to_ignorance",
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

Good feature — the FR↔EN alias map solves a real scoring gap and the test coverage is thorough (8 unit tests including integration with the scorer).

One minor concern on `_normalize_fallacy_type()`:

```python
if "_" in key:
    return key
return key
```

Both branches return `key` unchanged. The underscore check on line 68 is a no-op — the comment says it handles snake_case EN markers from the pipeline, but the function already handles that correctly via the alias-miss passthrough on the final `return key`. The dead branch is harmless but misleading; consider removing it or collapsing to a single `return key` to avoid future confusion.

The scorer change (iterating both `family` and `fallacy_type` fields) is correct and doubles the matching surface without false positives.

Rest is solid. Alias coverage looks comprehensive across the 8 families + yardstick markers.

Issue #953, opened 2026-06-05
Track FB-8 — Benchmark integral pipeline vs external-specialist yardstick (corpus_X) [Epic #947 Phase 4]


jsboigeEpita:
**Parent**: Epic #947 (Final Boss) — Phase 4 (spectacular-report DoD). **Lane**: po-2025 (heavy run). **BLOCKED / queued** behind Phase 0 (#941 + #945) AND #952 (the yardstick).

## Context (user mandate 2026-06-05, relayed via worker)

The "waouh" proof: run the integral pipeline on `corpus_X` and show its output is **comparable or superior** to a specialized external human analyst's published critique — a stronger baseline than 0-shot / journalistic. This is the terminal DoD of decision-B Epic #947, so it is **gated on trustworthy measurement**: do NOT run it until #941 (C1 extractor FOL `consistent` / Dung `extensions` fix) and #945 (integral re-run) have landed — otherwise the comparison inherits the under-reporting bug we are auditing out.

## Depends on
- #941 (C1 extractor trustworthy) — Phase 0
- #945 (integral re-run + regenerated comparison) — Phase 0
- #952 / FB-7 (external-specialist reference yardstick) — this Epic

## Goal
Score the integral pipeline output on `corpus_X` against the #952 yardstick: which analyst dimensions did the pipeline **match**, **miss**, or **EXCEED**? Produce the decision verdict (pipeline ≥ / ≈ / < specialist) with per-dimension evidence.

## DoD
- [ ] Integral run on `corpus_X` (trustworthy, post-#941/#945), state captured under the gitignored results path
- [ ] Per-dimension scorecard: pipeline output vs #952 yardstick (matched / missed / exceeded), each row backed by a text-ID + **a synthesis paragraph** (NO count-only table — synthesis-first mandate, incident R231)
- [ ] Explicit verdict: does the pipeline meet/beat the specialist? Where are the blind spots, and which are real-bug vs out-of-scope?
- [ ] Opaque IDs only; no source name / URL; downstream analysis under `argumentation_analysis/evaluation/results/` (gitignored)

## Privacy HARD
Opaque IDs only in ALL git artifacts. No `raw_text`/`full_text`. Consume the encrypted dataset in-memory via `load_extract_definitions`; never persist decrypted to disk.

## Anti-pendule
[...]

PR #999, opened 2026-06-07
fix(dung): graceful degradation on timeout with grounded-only fallback (#992)

jsboigeEpita:
## Summary

Fixes #992 — Dung extension computation hangs on large attack graphs (preferred/stable semantics), resulting in 0 extensions instead of a meaningful fallback.

### 
- Add  config (env , default 60s)
- Wrap  in 
- On : fall back to  (grounded-only)
- Add  +  to both timeout and exception fallback results
- Trace entry shows `DEGRADED — timeout` when applicable

### 
- `test_timeout_triggers_degraded_fallback`: mock hanging analyze_multi_semantics → verify timeout → grounded non-empty + degraded=true
- `test_exception_fallback_has_degraded_flag`: verify existing Exception path also sets degraded=true

## Impact on benchmark (#953)

In the corpus_X benchmark (BELOW verdict), D1/D8/D10 were all MISSED with `0 Dung extensions`. With this fix, the Dung phase will produce at least the grounded extension, enabling:
- D8 (Permission Architecture) — needs Dung framework output
- D2 (Functional Contradictions) — currently PARTIAL with 0 extensions, should improve
- Narrative synthesis has Dung data to work with

## Chain

#992 (this) → #993 (quality fallback) → #994 (narrative degradation) → re-run #953

## Files modified

| File | Type | Change |
|------|------|--------|
| `invoke_callables.py` | core | +52 lines (timeout wrapper + degraded flag) |
| `test_dung_aspic_wiring.py` | test | +86 lines (2 new tests)

### DIFF
--- a/argumentation_analysis/orchestration/invoke_callables.py
+++ b/argumentation_analysis/orchestration/invoke_callables.py
@@ -294,6 +294,12 @@ def llm_budget_scope(ceiling: Optional[int] = None) -> Iterator["_LLMBudget"]:
 # below a pathological hang. Set LLM_CALL_TIMEOUT_S=0 to disable.
 _LLM_CALL_TIMEOUT_S = float(os.environ.get("LLM_CALL_TIMEOUT_S", "300"))
 
+# Dung extension computation timeout (seconds). Preferred/stable semantics on
+# large attack graphs can hang indefinitely.  On timeout, falls back to
+# pure-Python grounded-only computation with degraded=True.  Set
+# DUNG_TIMEOUT_S=0 to disable timeout (not recommended).
+_DUNG_TIMEOUT_S = float(os.environ.get("DUNG_TIMEOUT_S", "60"))
+
 
 async def _guarded_chat_completion(client: Any, **kwargs: Any) -> Any:
     """Single funnel for every LLM chat-completion call (#708 runaway guard).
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

LGTM — clean timeout wrapper for the Dung extension computation.

**What works:**
- `asyncio.wait_for` + `asyncio.to_thread` is the correct pattern for timing out a blocking sync call in async context.
- Early `return` in the timeout path prevents the dangling `result` reference below — good.
- French trace entry matches existing conventions in the codebase.
- Default 60s is reasonable; `DUNG_TIMEOUT_S=0` opt-out documented.

**Minor nits (non-blocking):**
1. `os.environ.get("DUNG_TIMEOUT_S", "60")` passed to `float()` — if someone sets the env var to a non-numeric string (e.g. `DUNG_TIMEOUT_S=abc`), the module import crashes. Consider a try/except or a validation guard at module load time. Low priority since it is an operator config.
2. The `_fallback_result`, `_state`, `_n_args` local names use underscore prefix (typically reserved for module-level private symbols) — fine for local temporaries but worth noting for consistency.

**Tests:**
- `test_timeout_triggers_degraded_fallback` — properly mocks the hanging call, uses short timeout (0.3s), restores original in `finally`. Assertions cover `degraded`, `degraded_reason`, and grounded extension non-emptiness. Solid.
- `test_exception_fallback_has_degraded_flag` — clean, verifies the existing exception path also propagates the new flags.

No issues that would block merge.

PR #1000, opened 2026-06-07
fix(quality): catch OSError in _load_deps for torch DLL failure fallback (#993)

jsboigeEpita:
## Summary

Fixes #993 — Quality scoring returns all 0.0 when torch DLL fails (WinError 182) because the OSError propagates through spacy import and is not caught by the existing ImportError handler.

## Root Cause

The quality evaluator's  catches  but not . On Windows, when torch DLL fails (WinError 182), importing spacy triggers the DLL load, which raises . This propagates up to the detector-level  in , which catches it and sets all affected scores to 0.0.

### 
- Broaden  exception from  to 
- Add  flag to prevent repeated failing imports (was retrying on every detector call)
- Add  +  in evaluate() output when deps unavailable

### 
- : mock OSError → verify heuristic fallback produces non-zero scores
- : verify degraded=true + degraded_reason

## Impact on benchmark (#953)

In the corpus_X benchmark (BELOW verdict), all quality scores were 0.0. After fix:
- : 0.5-1.0 (avg word length heuristic)
- : 0.2-0.5 (substring count)
- : 1.0 (unique word ratio)
- Other virtues: 0.0 (French-specific patterns on English text — by design)

Average quality score goes from 0.0 to ~1.9/9. While still low for English text (detectors are French-optimized), it is no longer a flat zero.

## Chain

#992 (Dung, PR #999) → #993 (this) → #994 (narrative) → re-run #953

### DIFF
--- a/argumentation_analysis/agents/core/quality/quality_evaluator.py
+++ b/argumentation_analysis/agents/core/quality/quality_evaluator.py
@@ -27,11 +27,16 @@ _flesch_reading_ease = None
 _DEPS_AVAILABLE = False
 
 
+_DEPS_AVAILABLE = False
+_DEPS_ATTEMPTED = False
+
+
 def _load_deps():
     """Load optional dependencies (spacy, textstat) once."""
-    global _nlp, _flesch_reading_ease, _DEPS_AVAILABLE
-    if _DEPS_AVAILABLE:
-        return True
+    global _nlp, _flesch_reading_ease, _DEPS_AVAILABLE, _DEPS_ATTEMPTED
+    if _DEPS_ATTEMPTED:
+        return _DEPS_AVAILABLE
+    _DEPS_ATTEMPTED = True
     try:
         import spacy
         from textstat import flesch_reading_ease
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

The degradation logic is sound, but there is one bug to fix before merge.

**Bug — duplicate `_DEPS_AVAILABLE = False` declaration:**
Around line 27-30 of `quality_evaluator.py`, the patch introduces:

```
_DEPS_AVAILABLE = False   # original (line ~27)

_DEPS_AVAILABLE = False   # DUPLICATE — remove this
_DEPS_ATTEMPTED = False   # NEW — keep this
```

The second `_DEPS_AVAILABLE = False` is redundant. Functionally harmless (same value), but it is a code quality issue — likely a merge/edit artifact. Remove the duplicate line.

**What works:**
- `_DEPS_ATTEMPTED` guard is a good improvement — prevents re-attempting a doomed import on every call.
- Exception widening from `ImportError` to `(ImportError, OSError, RuntimeError)` is appropriate for the WinError 182 torch DLL propagation issue.
- `degraded`/`degraded_reason` fields in `evaluate()` output follow the same convention as PR #999.
- Log message updated to reflect the broader catch.

**Tests — acceptable with a caveat:**
- Both tests correctly reset global state (`_DEPS_ATTEMPTED`, `_DEPS_AVAILABLE`) and use `patch.dict(sys.modules)` to simulate the import failure.
- `test_dll_failure_returns_nonzero_scores` asserts specific score keys are > 0, which is reasonable for the given input.
- **Caveat:** The double reset of global state (once before `patch.dict`, once inside) is redundant — the inner one is sufficient since that is where `_load_deps` will be called. Not a bug, just unnecessary.

**Recommendation:** Fix the duplicate `_DEPS_AVAILABLE` line, then good to merge.

PR #1001, opened 2026-06-07
fix(narrative): graceful degradation - rebuild from context when state empty (#994)

jsboigeEpita:
## Summary

Fixes #994 — Narrative synthesis returns empty/fallback message even when pipeline has 14 args, 11 fallacies, 25 counter-arguments, 46 JTMS beliefs, and formal logic results.

## Root Cause

The state-based narrative reads from UnifiedAnalysisState fields, but at the time the narrative phase runs, the state may not yet be fully populated (timing/ordering issue in the pipeline). The build_narrative function returns the fallback sentinel message when state fields are empty.

### invoke_callables.py
- After state-based narrative, check if result is the fallback sentinel
- If so, read phase outputs directly from context dict and assemble a non-empty synthesis
- Covers: args count, fallacy types, counter-arguments, PL/FOL formulas, Dung extensions, JTMS beliefs, quality scores
- Add degraded=true + degraded_reason when rebuilding from context

### test_narrative_synthesis_spectacular.py
- test_narrative_from_context_when_state_empty: context with args+fallacies+CA -> non-empty narrative + degraded flag
- test_narrative_empty_context_still_returns: empty context -> no crash

## Chain

#992 (Dung, PR #999) -> #993 (Quality, PR #1000) -> #994 (this) -> re-run #953

### DIFF
--- a/argumentation_analysis/orchestration/invoke_callables.py
+++ b/argumentation_analysis/orchestration/invoke_callables.py
@@ -5733,6 +5733,10 @@ async def _invoke_narrative_synthesis(
     Reads state from context (populated by prior phases) and calls
     build_narrative to generate 1-2 paragraphs weaving together quality,
     fallacies, JTMS, ATMS, Dung, and formal logic results.
+
+    #994: Falls back to reading phase outputs directly from context when
+    the state-based narrative is empty/fallback, ensuring non-empty output
+    even when some upstream phases are degraded or missing.
     """
     from argumentation_analysis.plugins.narrative_synthesis_plugin import (
         build_narrative,
@@ -5740,6 +5744,11 @@ async def _invoke_narrative_synthesis(
     from argumentation_analysis.core.shared_state import UnifiedAnalysisState
     from argumentation_analysis.orchestration.state_writers import (
         CAPABILITY_STATE_WRITERS,
+    )
+
+    _FALLBACK_SENTINEL = (
+        "L'analyse n'a pas produit suffisamment de donnees pour generer "
+        "une synthese narrative"
     )
 
     # Reconstruct state from context if possible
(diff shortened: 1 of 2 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

## Review - fix(narrative): graceful degradation - rebuild from context when state empty (#994)

### Summary
Adds a fallback path in `_invoke_narrative_synthesis`: when `build_narrative(state)` returns empty or the hardcoded fallback sentinel, the function now reads phase outputs directly from `context` and assembles a degraded narrative from available data.

### Code quality - `invoke_callables.py`

**Positive:**
- Defensive `isinstance(extract_out, dict)` checks before `.get()` calls - safe against malformed context.
- `_FALLBACK_SENTINEL` is defined as a module-level constant, not a raw string - good practice.
- Result dict tagged with `degraded: True` and `degraded_reason` - downstream consumers can detect and handle degraded output.
- Logger info on degraded path helps debugging.
- No crash path: even with completely empty context, the function returns a valid dict with empty narrative.

**Minor observations:**
- The sentinel string is duplicated between here and wherever `build_narrative` produces it. If `build_narrative` changes its fallback message, this check silently breaks. Consider extracting the sentinel to a shared constant, or having `build_narrative` return `None`/empty string instead of a French sentence.
- `parts` list is joined with simple space concatenation. For readability, a newline or bullet format might be better for downstream display, but this is a style preference.

### Tests - `test_narrative_synthesis_spectacular.py`

**Positive:**
- Two test cases covering the happy path (context with data -> non-empty narrative) and edge case (empty context -> no crash).
- Asserts check narrative content (argument count, fallacy count, counter-arguments) and the `degraded` flag.
- Uses `pytest.mark.asyncio` correctly.

**Missing coverage:**
- No test for the sentinel detection path specifically (narrative containing the exact fallback message but non-empty). The current test only covers the `not narrative` branch. A test with `build_narrative` returning the sentinel would exercise the `_FALLBACK_SENTINEL in narrative` branch.
- No test verifying that when state IS valid, the degraded path is NOT taken (i.e. normal `build_narrative` result is returned unmodified).

**Verdict:** Clean implementation, no security issues, safe fallback. The missing sentinel-specific test is a gap but not a blocker.

PR #1012, opened 2026-06-08
docs(arch): multi-model routing design spec (#988)

jsboigeEpita:
## Summary -- #988 Multi-model routing design spec

Roadmap #78 Phase 2, North-Star parametric integration.

**Design spec only** — no code changes. invoke_callables.py is owned by po-2025 (collision guard).

### Design

- **Tier system**: cheap (18 mechanical phases) vs strong (11 reasoning phases)
- **Env vars**: LLM_MODEL_CHEAP / LLM_MODEL_STRONG — both default to OPENAI_CHAT_MODEL_ID (zero breaking change)
- **Implementation pattern**: extend _get_openai_client(tier=) and _guarded_chat_completion(tier=) with per-tier model resolution
- **Determinism**: per-model determinism params preserved (reasoning model suppression still works per resolved model)
- **Cost projection**: ~56% reduction on spectacular pipeline (29 phases → 18 cheap + 11 strong)

### Phase→Tier Mapping (26 phases)

| Phase | Tier | Rationale |
|-------|------|-----------|
| extract, nl_to_logic, text_to_kb, kb_to_tweety, quality | cheap | Mechanical/heuristic |
| pl, fol, modal, dung_extensions, aspic, jtms, atms, belief_revision, probabilistic, ranking, bipolar, tweety_interpretation, formal_synthesis, neural_detect | cheap | Algorithmic/Tweety |
| hierarchical_fallacy, counter, debate, governance, narrative_synthesis, analysis_synthesis, deep_synthesis, stakes | strong | Requires reasoning |

### Migration Path

1. **Phase 1**: Additive — add tier resolver, extend client/guard defaults (no caller changes)
2. **Phase 2**: Opt-in tier labels on mechanical phases
3. **Phase 3**: Cost-aware budget weighting

### Testing

5 unit tests + 1 integration test specified (Section 3).

### File

- docs/architecture/MULTI_MODEL_ROUTING_SPEC.md (new, 282 lines)

Closes: #988

### DIFF
--- a/docs/architecture/MULTI_MODEL_ROUTING_SPEC.md
+++ b/docs/architecture/MULTI_MODEL_ROUTING_SPEC.md
@@ -1,0 +1,282 @@
+# Multi-Model Routing Design Specification
+
+**Issue**: #988 — Roadmap #78 Phase 2, North-Star parametric integration
+**Scope**: Design spec only. No code changes to `invoke_callables.py` (collision guard: po-2025 owns that file's logic).
+**Purpose**: Define how per-phase model tier routing should work so that cheap mechanical phases (extraction, NL-to-logic) can use a cheaper model while reasoning-heavy phases (synthesis, governance, debate) use a strong model — reducing cost on repeated Phase 4 integral runs.
+
+**Status**: Design spec — awaiting implementation in a future sprint by the owner of `invoke_callables.py`.
+
+---
+
+## 1. Current State
+
+### 1.1 LLM Acquisition Path
+
+All LLM calls in the spectacular pipeline funnel through two functions in `invoke_callables.py`:
+
+| Function | Purpose | Location |
+|----------|---------|----------|
+| `_get_openai_client()` | Create `AsyncOpenAI` client + resolve model id | Line 175 |
+| `_get_determinism_params()` | Resolve temperature/seed (model-aware) | Line 127 |
+| `_resolve_model_id()` | Resolve active model id from env vars | Line 109 |
+| `_guarded_chat_completion()` | Budget-guarded chat completion funnel | Line 312 |
+
+**Current model resolution** (`_resolve_model_id()`):
+```
+If OPENROUTER_BASE_URL + OPENROUTER_API_KEY set:
+    model = OPENROUTER_CHAT_MODEL_ID || OPENAI_CHAT_MODEL_ID || "gpt-5-mini"
+Else:
+    model = OPENAI_CHAT_MODEL_ID || "gpt-5-mini"
+```
+
+**Key constraint**: `_resolve_model_id()` returns a **single** model id. Every phase in the pipeline uses this same model.
+
+### 1.2 Budget Guard
+
+`_guarded_chat_completion()` (line 312) funnels ALL LLM chat completions through a per-run budget. The circuit breaker is model-agnostic — it counts calls, not tokens or cost. Model-tier routing would allow the budget to become cost-aware (e.g., cheap calls count as 0.1, strong calls as 1.0).
+
+### 1.3 Determinism Logic
+
+`_get_determinism_params()` is already **model-aware**: it suppresses temperature/seed for reasoning models (gpt-5*, o1*, o3*) unless `LLM_FORCE_SAMPLING_PARAMS=1`. This logic MUST be preserved per resolved model (not just the global model).
+
+---
+
+## 2. Proposed Architecture
+
+### 2.1 Tier System
+
+Introduce a **tier** abstraction over model ids:
+
+| Tier | Purpose | Default resolution | Example models |
+|------|---------|-------------------|----------------|
+| `cheap` | Mechanical phases (extraction, parsing, NL-to-logic) | Falls back to `OPENAI_CHAT_MODEL_ID` | `gpt-5-mini`, `gpt-4.1-mini` |
+| `strong` | Reasoning-heavy phases (synthesis, debate, governance) | Falls back to `OPENAI_CHAT_MODEL_ID` | `gpt-5-mini`, `gpt-4.1`, `o3-mini` |
+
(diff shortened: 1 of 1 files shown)


jsboigeEpita (commented):
**[NanoClaw]**

Well-structured design spec. Detailed review below.

## Strengths

- **Zero-breaking-change default**: both tiers falling back to `OPENAI_CHAT_MODEL_ID` is the right call. Existing deployments are unaffected until someone explicitly sets `LLM_MODEL_CHEAP` or `LLM_MODEL_STRONG`.
- **Collision guard**: explicitly scoping implementation to po-2025 prevents merge conflicts on `invoke_callables.py`.
- **Determinism awareness**: correctly identifying that `_get_determinism_params()` needs per-model resolution (not just per-tier) is a subtle but important detail.
- **Phase-to-tier mapping**: the 18 cheap / 11 strong split is conservative and defensible. Algorithmic phases (pl, fol, modal, jtms, atms, dung) correctly identified as not needing LLM reasoning power.
- **Phased migration**: additive approach (Phase 1 extends signatures with defaults, Phase 2 opts in) is the right way to avoid a big-bang change.

## Concerns

### 1. Inconsistency: phase count mismatch

Section 2.3 header says "29 phases" and the cost projection uses 29. But the Phase-to-Tier Mapping table only lists **26 rows**. Three phases are missing. This needs reconciliation — either the table is incomplete or the count is wrong. A future implementor will wonder which 3 phases to classify.

### 2. Cost projection uses arbitrary weight (0.1)

Section 2.3 calculates `18 x 0.1 + 11 x 1.0 = 12.8`. The 0.1 weight for cheap tiers is stated without justification. In Section 4, the actual per-token cost difference between `gpt-5-mini` ($0.15/$0.60) and `gpt-4.1-mini` ($0.10/$0.40) is only ~33% cheaper, not 10x. Using 0.1x inflates the projected savings. I'd suggest either (a) using actual token cost ratios, or (b) clearly labeling 0.1 as a conceptual weight, not a cost ratio.

### 3. `stakes_extraction` classified as `strong`

Section 2.3 marks `stakes` as `strong` with rationale "Requires stakeholder analysis reasoning." However, `stakes_extraction` (like `fact_extraction`) is primarily a structured extraction task — identifying stakeholders and their positions from text. Consider whether this should be `cheap` unless the implementation actually does generative reasoning beyond extraction.

### 4. OpenRouter resolution priority asymmetry

The OpenRouter resolution chain in Section 2.2 is:
```
OPENROUTER_MODEL_CHEAP || LLM_MODEL_CHEAP || OPENROUTER_CHAT_MODEL_ID || OPENAI_CHAT_MODEL_ID
```
But in the existing codebase (Section 1.1), the resolution is:
```
OPENROUTER_CHAT_MODEL_ID || OPENAI_CHAT_MODEL_ID
```

The new chain inserts `LLM_MODEL_CHEAP` between `OPENROUTER_MODEL_CHEAP` and `OPENROUTER_CHAT_MODEL_ID`. This means a non-OpenRouter env var (`LLM_MODEL_CHEAP`) would override an OpenRouter-specific one (`OPENROUTER_CHAT_MODEL_ID`). Is this intentional? The existing pattern is that OpenRouter vars take precedence when OpenRouter is active.

### 5. Missing: error handling for invalid tier

The implementation sketch in Section 2.4A returns `tier_model = None` for unrecognized tiers, which falls back to `base_model`. This silently ignores typos (e.g., `tier="cehap"`). Consider logging a warning for unknown tier values, or using an enum/constant.

### 6. Testing gap: OpenRouter path

The unit tests in Section 3.1 only cover the OpenAI resolution path. There should be a test for the OpenRouter priority chain (e.g., `OPENROUTER_MODEL_CHEAP` overriding `LLM_MODEL_CHEAP` when OpenRouter is active).

### 7. Migration Phase 3 is underspecified

Section 5.3 mentions extending the budget guard to be cost-aware (`cheap = 0.1x, strong = 1.0x`) and a `--llm-budget` CLI param. But there's no design for how the weighting would work with the existing call-count-based circuit breaker. Is it a separate budget counter? Does the circuit breaker threshold change? This phase needs more detail before implementation.

## Minor Notes

- Section 1.1 says "two functions" but lists four. Minor wording fix.
- `neural_detect` is marked `cheap` because it uses an NLP model inference — worth noting that if the NLP model is a local spaCy/textstat model rather than an LLM call, it doesn't go through `_guarded_chat_completion` at all, so tier routing would be a no-op for it.
- The cross-references in Section 7 are good but would benefit from line numbers or function signatures for easier lookup.

## Verdict

Solid design doc with a clear migration path. The main issues to address before implementation: (1) reconcile the phase count (26 vs 29), (2) fix the cost projection arithmetic or clarify the weight semantics, and (3) decide on the OpenRouter priority chain ordering.

clusterManager-Myia (commented):
### [Hermes] Review — COMMENT_WITH_CONCERNS

**Checklist:**
- ✅ Security scan: CLEAN (design spec only, no code)
- ✅ Cross-repo impact: aucun (design doc only)
- ✅ Notebooks: N/A
- ✅ Fix/incident: N/A

**Review:** Design spec multi-model routing (#988). Architecture claire: tier system (cheap/strong), env vars LLM_MODEL_CHEAP/STRONG. Points de vigilance:
1. **Fallback strategy**: Que se passe-t-il si LLM_MODEL_CHEAP est indisponible ? Le spec mentionne 'both default to OPENAI_CHAT_MODEL_ID' mais pas de cascade cheap→strong en cas d'erreur.
2. **Latency budget**: Pas de budget de latence défini pour le threshold cheap→strong. Risque de fallback systématique si le cheap est trop lent.
3. **invoke_callables.py ownership**: Bonne garde (po-2025). Mais le spec devrait lister les fonctions qui seront affectées par le routing.

**Decision:** COMMENT_WITH_CONCERNS — spec solide mais 3 points à clarifier avant implementation.

PR #1030, opened 2026-06-09
fix(budget): wire LLM budget guard into SK-native conversational path (#1029)

jsboigeEpita:
## Summary

FB-5 #1029 — B-F, last hole in #947 Phase 2.

The budget scope (`llm_budget_scope`) was entered on the conversational path, but SK-native calls (`AgentGroupChat.invoke` / `ChatCompletionAgent.invoke`) bypass `_guarded_chat_completion` entirely — the counter never incremented, making the circuit-breaker **inert** for the entire conversational dialogue. This is the exact path of the 8h / 12.4K-call runaway incident #708.

## Root Cause

```
Sequential path:  _guarded_chat_completion() ← counter increments ✓
Conversational:   chat.invoke() → SK SDK → API ← counter stays at 0 ✗
```

Post-processing calls (counter-arg, quality, formal logic) were already covered — they go through `_guarded_chat_completion`. The gap was exclusively the SK-native dialogue turns.

## Fix

- Add `_bump_sk_budget()` helper that reads the ContextVar and increments the shared counter
- Wire it into **all 5 SK-native invoke sites**:
  - `conversational_orchestrator.py`: 4 sites (chat.invoke ×2, agent.invoke ×2)
  - `conversational_executor.py`: 1 site (chat.invoke)
- Raises `LLMBudgetExceeded` when ceiling exceeded (same exception as pipeline path)

## Tests (9 new)

`test_conversational_sk_budget.py`:
- `_bump_sk_budget` unit: noop/no-budget, increment, ceiling, shared counter
- `_bump_sk_budget_exec` unit: increment, ceiling
- Integration: round-robin path bumps budget per turn, runaway fires cap
- Non-regression: budget scope is entered

## Collision Guard

- ✅ Does NOT touch `quality_evaluator.py` (po-2025)
- ✅ Does NOT touch benchmark harness (#1027 po-2025)

## Verification

```
9 passed, 9 warnings in 2.30s
```

Closes #1029
Refs: #947 Phase 2, #708 (origin incident), #1019

### DIFF
--- a/argumentation_analysis/orchestration/conversational_executor.py
+++ b/argumentation_analysis/orchestration/conversational_executor.py
@@ -36,23 +36,7 @@ from argumentation_analysis.orchestration.workflow_dsl import (
 )
 
 
-def _bump_sk_budget_exec(n: int = 1) -> None:
-    """Increment the LLM budget counter for SK-native agent calls (#1029)."""
-    from argumentation_analysis.orchestration.invoke_callables import _llm_budget
-
-    budget = _llm_budget.get()
-    if budget is not None:
-        budget.count += n
-        if budget.count > budget.ceiling:
-            from argumentation_analysis.orchestration.invoke_callables import (
-                LLMBudgetExceeded,
-            )
-
-            raise LLMBudgetExceeded(
-                f"Global LLM-call budget exceeded "
-                f"({budget.count} > {budget.ceiling}) in conversational pipeline "
-                f"— runaway guard (issue #708)."
-            )
+from argumentation_analysis.orchestration.invoke_callables import _bump_sk_budget
 
 
 logger = logging.getLogger("ConversationalExecutor")
@@ -440,7 +424,7 @@ class GroupChatTurnStrategy(TurnStrategy):
             # Collect messages from the async generator
             collected: List[ChatMessageContent] = []
             async for msg in chat.invoke():
-                _bump_sk_budget_exec()
+                _bump_sk_budget()
                 collected.append(msg)
 
             logger.info(
--- a/argumentation_analysis/orchestration/conversational_executor.py
+++ b/argumentation_analysis/orchestration/conversational_executor.py
@@ -35,6 +35,26 @@ from argumentation_analysis.orchestration.workflow_dsl import (
     WorkflowExecutor,
 )
 
+
+def _bump_sk_budget_exec(n: int = 1) -> None:
+    """Increment the LLM budget counter for SK-native agent calls (#1029)."""
+    from argumentation_analysis.orchestration.invoke_callables import _llm_budget
+
+    budget = _llm_budget.get()
+    if budget is not None:
+        budget.count += n
+        if budget.count > budget.ceiling:
+            from argumentation_analysis.orchestration.invoke_callables import (
+                LLMBudgetExceeded,
+            )
+
+            raise LLMBudgetExceeded(
+                f"Global LLM-call budget exceeded "
+                f"({budget.count} > {budget.ceiling}) in conversational pipeline "
+                f"— runaway guard (issue #708)."
+            )
+
+
 logger = logging.getLogger("ConversationalExecutor")
 
 
@@ -420,6 +440,7 @@ class GroupChatTurnStrategy(TurnStrategy):
             # Collect messages from the async generator
             collected: List[ChatMessageContent] = []
             async for msg in chat.invoke():
+                _bump_sk_budget_exec()
                 collected.append(msg)
 
             logger.info(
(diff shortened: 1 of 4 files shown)


clusterManager-Myia (commented):
**[NanoClaw]**

## Summary

The PR wires the existing `_LLMBudget` circuit-breaker into the SK-native conversational paths (both `AgentGroupChat.invoke()` and `ChatCompletionAgent.invoke()` round-robin fallback), closing a gap where SK's direct OpenAI calls bypassed `_guarded_chat_completion`. Good defensive addition per incident #708.

## Deep review -- logic correctness

### 1. Per-chunk vs per-turn granularity (design concern)

The budget bump is placed **inside** every `async for` loop body:

```python
async for response in agent.invoke(chat_history):  # round-robin
    _bump_sk_budget()  # fires per yielded chunk
    ...
```

- For `AgentGroupChat.invoke()`: the async generator yields one `ChatMessageContent` per agent turn, so the bump is correctly **per-turn**.
- For `ChatCompletionAgent.invoke()` (round-robin fallback at L1496, L1534): the SK docs indicate this can yield **multiple streaming chunks per invocation**. The bump would fire per-chunk rather than per-LLM-call, inflating the counter and making the ceiling unpredictable.

**Recommendation**: For the round-robin path, move `_bump_sk_budget()` **before** the `async for` loop (once per turn) rather than inside it. This matches the semantics of one LLM API call = one budget tick regardless of streaming behavior.

### 2. Exception caught per-turn neutralizes the guard (concern)

In `_run_phase` round-robin path (L1554), the loop body is wrapped in `try/except Exception`. When `_bump_sk_budget()` raises `LLMBudgetExceeded` mid-iteration of the `async for` generator, the exception is caught, logged as an error, and the loop continues to the next turn. The budget counter keeps incrementing but never actually stops execution.

The test `test_budget_cap_fires_on_runaway` (L183) acknowledges this: it asserts `budget.count > budget.ceiling` but does **not** assert that the loop stopped early. The comment says _run_phase catches per-turn exceptions -- this is correct behavior for the current code, but it means the guard is informational (logged) rather than functional (pipeline-stopping) on this path.

The existing `_guarded_chat_completion` path has the same property (the budget exception is caught per-phase), so this is consistent -- but worth noting that the runaway guard effectively becomes a runaway logger on the conversational path.

### 3. Duplicate functions (minor)

`_bump_sk_budget()` in `conversational_orchestrator.py` and `_bump_sk_budget_exec()` in `conversational_executor.py` are nearly identical (3-line difference: function name, docstring length, error message suffix pipeline vs phase). Consider extracting a shared utility to avoid drift -- e.g. `_increment_llm_budget()` in `invoke_callables.py` alongside the existing budget infrastructure.

### 4. Tests -- good coverage, one caveat

The unit tests for the counter mechanism (TestBumpSKBudget / TestBumpSKBudgetExecutor) are thorough and correctly test no-op, increment, ceiling enforcement, and ContextVar sharing.

The integration test `test_round_robin_path_bumps_budget` uses `yield 100` chunks per invoke to prove the bump fires. However, given concern #1 above, this test proves the per-chunk behavior exists but does not validate that it is the *desired* behavior.

`test_budget_cap_fires_on_runaway` correctly proves the exception is raised, but as noted in #2, does not prove the pipeline stops.

## Security scan

No secrets, tokens, credentials, or private keys found in the diff. Clean.

## Verdict

The budget guard wiring is **structurally correct** for `AgentGroupChat` (per-turn), but has a **granularity mismatch** for `ChatCompletionAgent` streaming (per-chunk) and the exception is effectively caught and swallowed per-turn in the round-robin path. Both are real but non-blocking -- the PR still provides meaningful runaway detection and logging. Recommend addressing the per-chunk issue before merge to avoid false ceiling triggers in production.

jsboigeEpita (commented):
**[Hermes] LGTM** — Approve-equivalent (self-review cap).

**Checklist:**
1. ✅ Security scan: CLEAN
2. ✅ Cross-repo: Epita-only, no collision with CoursIA/po-2025
3. ✅ NC concerns (e58893ef → 1556e259): ALL ADDRESSED
   - **#1 per-chunk**: bump moved BEFORE `async for` at both `ChatCompletionAgent.invoke()` sites (L1475, L1516)
   - **#2 exception swallowed**: `except LLMBudgetExceeded: raise` before generic `except Exception` (anti-theater #1019)
   - **#3 duplicate functions**: single `_bump_sk_budget()` in `invoke_callables.py`
4. ✅ Tests: per-call granularity verified (3 turns × 5 chunks = 3, not 15). Anti-regression confirms exception propagates.

**Verdict:** All NC concerns resolved. Ready to merge — needs 2nd reviewer for green light.

PR #1060, opened 2026-06-11
fix(orchestration): wire DelegatingSelectionStrategy into SK-native AgentGroupChat (#1051)

jsboigeEpita:
### Problem

The PM calls \designate_next_agent()\ via StateManagerPlugin — state IS modified — but on the primary SK-native path, \AgentGroupChat(agents=agents)\ was instantiated **without** \selection_strategy\. SK's internal round-robin never reads \_next_agent_designated\. Net effect: PM directs into the void.

### Fix

1. **Wire DelegatingSelectionStrategy** into _run_phase's AgentGroupChat instantiation
2. **Document inline PM** as canonical for conversational path
3. **5 regression tests** proving designation is consumed

### Test Results

5 passed (test_ra6_pm_designation.py)
18 passed (test_strategies.py — no regression)

### Anti-pendule

- Round-robin fallback path preserved
- No changes to pm_agent.py or enhanced_pm_analysis_runner.py

Closes #1051

### DIFF
--- a/argumentation_analysis/orchestration/conversational_orchestrator.py
+++ b/argumentation_analysis/orchestration/conversational_orchestrator.py
@@ -151,6 +151,12 @@ def _detect_language(text: str) -> str:
 # See ARCHEOLOGIE_ORCHESTRATION.md section 3 for rationale.
 # ---------------------------------------------------------------------------
 
+# NOTE (RA-6 #1051): The inline "ProjectManager" ChatCompletionAgent defined below
+# is the CANONICAL PM for the conversational path. It uses StateManagerPlugin's
+# designate_next_agent() kernel function, which writes to state._next_agent_designated.
+# DelegatingSelectionStrategy (wired in _run_phase) reads that field to honour PM
+# designation. The separate ProjectManagerAgent in agents/core/pm/pm_agent.py serves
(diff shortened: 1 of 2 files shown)


jsboigeEpita (commented):
**[Hermes]** Review COMMENT

## Summary
Good fix: wires DelegatingSelectionStrategy into the SK-native AgentGroupChat path so PM designation is actually consumed.

## Production code (conversational_orchestrator.py, +29/-1)
- ✅ Defensive import (try/except ImportError + TypeError/ValueError) — degrades to SK default gracefully
- ✅ Guard on `state is not None and hasattr(state, "consume_next_agent_designation")` — safe
- ✅ Clear NOTE comment explaining why the inline PM differs from the one in agents/core/pm/
- No secrets found

## Tests (test_ra6_pm_designation.py, +210)
- ✅ 3 test classes covering strategy behaviour, wiring, and integration
- ✅ Uses mock agents (no API keys needed)
- ✅ Tests designation consumption (no double-read leak) — the key regression case
- ✅ Tests default fallback when no designation set

## Minor note
`asyncio.run(strategy.next(...))` in tests is fine for synchronous pytest, but if these ever move to an async test runner, they would need `pytest-asyncio`. Not blocking for now.

PR #1092, opened 2026-06-14
docs(fb-24): #1090 WSL execution path for the quality radar — proven

jsboigeEpita:
## FB-24 #1090 — WSL execution path for the quality radar (parallel to #1088)

**Dispatch**: R398 from `myia-ai-01` (FB-24 #1090)
**Base**: `main` `7019b620` (post #1089)
**Author**: Claude Code @ `myia-po-2025:2025-Epita-Intelligence-Symbolique`

### Problem

After FB-22 (#1089) lifted the formal block to **DECIDES**, the Epic #947 verdict is capped at EDGES by **exactly one thing**: the quality category is unevaluated because the quality radar crashes on Windows-native with spaCy/torch `WinError 182` (`fbgemm.dll` DLL fault, issue #882). User mandate (2026-06-14): native crashes get **fixed or moved to WSL** — never documented-and-accepted.

This PR operationalizes the **WSL fallback path** in parallel with po-2023's Windows-native fix (#1088).

### Result — PROVEN

| Check | Result |
|-------|--------|
| spaCy + torch load (torch-before-spacy, conftest order) | ✅ `spacy 3.8.14 \| torch 2.6.0 \| platform=linux` |
| spaCy + torch load (spacy-before-torch, the Windows-fatal order) | ✅ loads — order irrelevant under Linux |
| `WinError 182` | **ABSENT** under WSL (no Windows DLL conflict) |
| `fr_core_news_sm` model | ✅ 3.8.0 |
| Quality dimension on **corpus_C** | ✅ note_finale 2.0/9, **Flesch 46.34** (real textstat), 3/9 non-zero virtues |
| Quality dimension on **corpus_A** | ✅ **Flesch 40.12**, 3/9 non-zero virtues |

**This is the answer to #1088's native-vs-WSL question**: WSL works; the DLL conflict is Windows-only.

### DoD (#1090)

- [x] spaCy + torch load under WSL with NO `WinError 182` (captured proof, both import orders)
- [x] Quality dimension produces real non-trivial values on ≥1 corpus (done on 2: C and A)
- [ ] WSL env runs the FULL harness (conda+deps ✅, JDK/JPype startJVM ⏳ — sudo-gated; **the quality dimension needs no JVM**, so this is only relevant if a future full-pipeline run wants WSL environmental consistency). Documented as the one completion step.
- [x] WSL invocation documented (reproducible + reusable by the A/B/C quality re-measurement) → `docs/guides/WSL_QUALITY_RADAR.md`

### Findings

[...]

PR #1091, opened 2026-06-14
fix(quality): provision real quality-radar deps — refuted WinError 182 (#1088)

jsboigeEpita:
FB-23 #1088 (Epic #947 Final Boss). Quality dimension FAILED on every corpus, attributed to WinError 182. Fresh-process repro REFUTED that: spaCy + fr_core_news_sm load cleanly even with jpype-first ordering. Verified root cause = missing deps (textstat not in any spec; fr_core_news_sm never downloaded by setup). Durable fix: textstat in environment.yml+requirements.txt, spacy download step in setup_project_env.ps1, reworded error messages, recipe-doc correction note, + fix latent inter-test pollution (autouse fixture resets _DEPS_ATTEMPTED). Verified on projet-is: _invoke_quality_evaluator real path aggregate_score=3.3, 3 args scored non-trivially; quality dir 53 passed/0 failed (no regression). WinError 182 NOT reproducible. Out of scope: test_llm_enrich_quality (8 pre-existing fails on origin/main). Unblocks po-2025 #1087 quality numbers on Windows-native (WSL now optional). Closes #1088. See full body in commit f3c856fd.

PR #1095, opened 2026-06-14
fix(quality): neutralize faulty torch so rule-based spaCy model loads natively (FB-25 #1093)

jsboigeEpita:
## What & why (FB-25 #1093, Epic #947 final gate — item 1)

On Windows envs where `torch` is installed but its native DLL faults (`WinError 182` / `fbgemm.dll`, issue #882 — e.g. torch 2.2.2 in `projet-is-roo-new`), spaCy's `import spacy` pulls thinc, which does an *optional* `import torch`. thinc catches `ImportError` but **not** `OSError` — so the DLL fault propagates and blocks spaCy entirely, even though the `fr_core_news_sm` model is **rule-based (non-neural) and does not need torch**. The quality radar therefore fails loud on these envs (fail-loud gate #1019 is respected, but the radar is unusable).

This was the reconciliation surfaced in R399 between #1091 (deps fix, `projet-is`) and #1092 (WSL path): the diagnosis is **env-specific**. `projet-is-roo-new` has a broken torch, so #1091's deps fix alone does NOT unblock it there. This PR unblocks the native path in broken-torch envs without an env rebuild.

## Fix (anti-pendule: remove the poisoning, no counterweight)

`_neutralize_faulty_torch()` (new, typed `-> None`) runs inside `_load_deps()` **before** `import spacy`:
- try `import torch`; if it raises **OSError** (DLL fault) → plant a `None` sentinel in `sys.modules["torch"]`. A subsequent `import torch` then raises `ImportError` ("import of name halted; None in sys.modules"), which thinc catches gracefully and skips → spaCy loads the rule-based model.
- **no-op** when torch is genuinely absent (`ImportError`, thinc already handles) or imports cleanly (neural path stays available — critical so the wide-net / camembert path is unaffected in healthy-torch envs like `projet-is`).

It only fires when torch is installed-but-broken, in which case the neural path was never usable in this process anyway. Idempotent (`_TORCH_NEUTRALIZED` guard).

## Proof (native `projet-is-roo-new`, helper-driven — no manual `sys.modules` hack)

Real `ArgumentQualityEvaluator.evaluate` on `corpus_C` (4000-char opaque excerpt, dataset in-memory):

```text
note_finale: 2.0 | note_moyenne: 0.222 | non-zero virtues: 3/9
VERDICT: NATIVE quality works via integrated helper ✅
```

Values are **identical to WSL (#1092) and the R399 manual probe** (note_finale 2.0, Flesch 46.34) — the radar is deterministic; the fix only changes whether it can run at all on broken-torch envs.

## Tests

- New `TestNeutralizeFaultyTorch` (3 cases): faulty→blocked, absent→noop, healthy→noop. **FB-21 lesson**: the bug shipped because nothing asserted this dispatch.
- Full quality dir: **56 passed, 0 failed**. Existing `TestDllFailureFailLoud::test_dll_failure_raises_runtime_error` still passes (it patches spacy/textstat directly, independent of torch — my helper runs first, neutralizes the real broken torch, then the patched-spacy ImportError still fires the expected RuntimeError).
- mypy strict on the file: **0 new errors** (7 pre-existing on origin/main, all about untyped `_load_deps` — unchanged).

## Scope / privacy

2 files: `quality_evaluator.py` (+ helper + call) + test file (+ regression class). File-disjoint from FB-26 #1094 (po-2023, audit doc + verdict section). Privacy HARD: opaque corpus IDs, aggregate-only output, dataset consumed in-memory. No heuristic fallback added (#1019 honored — real spaCy/textstat values).

Closes the native half of the Epic #947 quality gate; combined with the WSL path (#1092 merged) the radar is now runnable everywhere. Item 2 of FB-25 (A/B/C re-measure + report fold) follows in a separate commit, serialized with po-2023 on `FB20_PHASE4_TERMINAL_REPORT.md`.

### DIFF
--- a/argumentation_analysis/agents/core/quality/quality_evaluator.py
+++ b/argumentation_analysis/agents/core/quality/quality_evaluator.py
@@ -23,6 +23,7 @@ import json
 import logging
 import os
 import re
+import sys
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple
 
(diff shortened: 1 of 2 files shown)


jsboigeEpita (commented):
**Review COMMENT_WITH_CONCERNS** (self-review cap → COMMENT uniquement)

**Points positifs :**
- Diagnostic correct et fix anti-pendule (neutraliser le poison, pas ajouter un contrepoids).
- Idempotence via `_TORCH_NEUTRALIZED` guard.
- 3 tests de régression couvrant les 3 branches (faulty/absent/healthy).
- `noqa: F401` sur le test import torch expliqué par le test structurel.

**Concerns :**
1. **`sys.modules["torch"] = None` est un monkey-patch global** — si un autre module dans le même process importe torch *après* la neutralisation mais *avant* spacy, il sera aussi bloqué. Acceptable ici (quality path = premier importeur de torch) mais值得 un commentaire inline warning dans `_neutralize_faulty_torch` que c'est un process-wide side-effect, pas un scope-local.
2. **La teardown dans les tests** (`sys.modules.pop("torch", None)`) ne restaure pas `builtins.__import__` (elle est patchée via context manager, donc restaurée automatiquement par `patch()` — OK). Mais le test `test_healthy_torch_is_not_blocked` fait `sys.modules.pop("torch", None)` *avant* le patch, ce qui est correct mais non-obvious — un `# ensure clean slate` commentaire aiderait.

Aucun concern bloquant. Fix correct pour le problème réel (issue #882).

jsboigeEpita (commented):
[Hermes] **COMMENT** — clean, well-tested fix for the broken-torch DLL fault blocking spaCy.

**Checklist:**
- ✅ Security scan: no tokens/secrets
- ✅ Root cause: torch 2.2.2 DLL fault (WinError 182) propagates via thinc (catches ImportError, not OSError)
- ✅ Fix: plant `None` sentinel in `sys.modules["torch"]` → ImportError (which thinc handles). No-op when torch absent/healthy.
- ✅ Anti-pendule: removes the poisoning, doesn't add a counterweight
- ✅ Tests: 3 regression guards (faulty blocked, absent untouched, healthy untouched) + clean teardown

Solid fix. Would APPROVE if not self-review.

PR #1154, opened 2026-06-17
fix(restitution): #1153 act2 reader↔writer schema (quality virtues, PL verdict, Dung label, non-attributed, usable gate)

jsboigeEpita:
Closes #1153.

## Summary

Implements the 5 reader-side fixes proven in audit #1151 (#1152), **all in `act2_narrative_plugin.py`** — file-disjoint from R4's `act3_conclusion_plugin.py`, runs in parallel with po-2025's R4. **Zero edits to the 5 infra files** (`shared_state`, `invoke_callables`, `registry_setup`, `state_writers`, `workflows`) — R4's lane.

Validation: **110 restitution tests pass** (0 regression), **mypy strict clean**. 12 new contract tests.

## Fixes (5)

| # | Fix | Writer key (file:l) | Reader now | Severity |
|---|-----|---------------------|------------|----------|
| **A** | quality virtues | `scores` shared_state.py:536 | `qs.get("scores")` + legacy `scores_par_vertu` | HIGH |
| **C** | PL verdict | `satisfiable` shared_state.py:801 | `_pl_verdict()` reads `satisfiable` + legacy `consistent` | HIGH |
| **D** | Dung semantics | (none; in `name`) state_writers.py:717 | parsed from `name=verification_<sem>` | LOW |
| **(a)** | non-attributed fallacy | target_arg_id unresolved | counted in `Act2Evidence.unattributed_fallacies` (was silently dropped) | MED |
| **§5** | usable-content gate | — | `_quality_axis_usable()` replaces `bool(dict)` | MED |

### Finding C — strict `is True`/`is False` preserved (#1019)
```python
def _pl_verdict(result):
    sat = result.get("satisfiable")
    if sat is None:
        sat = result.get("consistent")
    return sat if isinstance(sat, bool) else None  # NEVER bool(sat)
```
A `None` (unverified) verdict returns `None`, never collapses to `False` (inconsistent) — anti-#1019.

### §5 gate — calibrated, not over-strict (anti-pendule)
`_quality_axis_usable()` requires **at least one** entry with a usable `overall` OR per-virtue map. A single overall-only entry (legitimate partial state) stays available — no false fail-loud. An empty-but-present entry (`{"arg_1": {}}`) correctly fails loud (this is where Finding A hid).

### Why the bug was invisible
[...]

PR #1158, opened 2026-06-18
fix(restitution): #1151 §3(b) stakes extractor argument schema

jsboigeEpita:
## What

`_invoke_stakes_extractor` read `list(getattr(state, "identified_arguments", []))`. Since `identified_arguments` is a **dict** `{arg_id: description}` (writer schema, [shared_state.py:103](argumentation_analysis/core/shared_state.py#L103)), `list(dict)` passed **bare arg_id keys** (strings) to `StakesExtractor.extract` — which expects `List[Dict]` with a `"text"`/`"description"` key and calls `arg.get("text", ...)` on each. On real (LLM-on) runs that's an **AttributeError** crashing stakes extraction.

## Why it was invisible

Unit tests mock `LLM=None`; `extract` short-circuits to an empty result **before** the arguments loop — so the crash never fired in tests, only in real runs. Same *"fixture encodes the consumer's expected contract, not the writer's real schema"* pattern as Finding A/C/D (#1154).

This is **§3(b)** from po-2023's audit #1151 — assigned to po-2025's lane (R432) and folded here as promised.

## Fix (anti-pendule)

Convert the dict-of-strings into the list-of-dicts the consumer's contract specifies, at the single call site — feed the consumer what it asked for, no counterweight. Defensive fallback for a sequence-of-dicts (legacy/fixture) preserved.

```python
raw_args = getattr(state, "identified_arguments", {}) or {}
if isinstance(raw_args, dict):
    arguments = [{"text": desc} for desc in raw_args.values() if isinstance(desc, str)]
else:  # defensive: already a sequence of argument dicts (legacy/fixture)
    arguments = list(raw_args)
```

## Tests

- Realigned the 2 stakes fixtures to the writer schema (`{arg_id: desc}`).
- **Added `test_dict_arguments_passed_as_list_of_dicts`**: pins the contract by capturing what `extract` receives (bypasses the LLM=None short-circuit that masked the bug) — asserts a list of dicts each with `"text"`, real descriptions present, no bare arg_id keys leaking.

## Validation

- **mypy strict** (`invoke_callables.py` = CI-gated): **Success, 0 issues**
- mute_capabilities + stakes tests: **26 passed**
- Restitution suite: **134 passed** (0 regression)
- **0 LLM spend**

Refs #1151. File-disjoint from #1154 (act2) and R5-v2 (act plugins) — `invoke_callables.py` is po-2025's infra lane.

### DIFF
--- a/argumentation_analysis/orchestration/invoke_callables.py
+++ b/argumentation_analysis/orchestration/invoke_callables.py
@@ -6857,8 +6857,23 @@ async def _invoke_stakes_extractor(
 
     extractor = StakesExtractor()
 
-    # Gather arguments from state
-    arguments = list(getattr(state, "identified_arguments", []) or [])
+    # Gather arguments from state.
+    # Writer schema (shared_state.add_argument): identified_arguments is a
+    # dict {arg_id: description}. StakesExtractor.extract expects a list of
+    # argument dicts with a "text"/"description" key. The previous
+    # ``list(dict)`` passed bare arg_id keys (strings), which crashed
+    # ``arg.get("text", ...)`` inside extract on real (LLM-on) runs — masked
+    # by unit tests that mock LLM=None (extract short-circuits before the
+    # loop). Audit #1151 §3(b). Convert the dict-of-strings into the
+    # list-of-dicts the consumer's contract specifies (anti-pendule: feed the
+    # consumer what it asked for, no counterweight).
+    raw_args = getattr(state, "identified_arguments", {}) or {}
+    if isinstance(raw_args, dict):
+        arguments = [
+            {"text": desc} for desc in raw_args.values() if isinstance(desc, str)
+        ]
+    else:  # defensive: already a sequence of argument dicts (legacy/fixture)
+        arguments = list(raw_args)
 
     # Source metadata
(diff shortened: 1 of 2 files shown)


jsboigeEpita (commented):
**[Hermes]** COMMENT — clean bug fix.

**Checklist:**
- ✅ Security scan: no tokens/secrets in diff.
- ✅ Root cause clear: `list(dict)` passed bare keys instead of `{\"text\": desc}` dicts to StakesExtractor.extract. Crash masked by LLM=None mock short-circuit in tests.
- ✅ Fix is minimal and anti-pendule: single call site, defensive fallback for legacy/fixture sequences preserved.
- ✅ New test `test_dict_arguments_passed_as_list_of_dicts` pins the contract by stubbing extract and asserting received structure — correctly bypasses the LLM=None short-circuit that masked the bug.
- ✅ Existing fixtures realigned to writer schema `{arg_id: desc}`.
- ✅ mypy strict clean, 26 stakes tests pass, 134 restitution suite pass.
- ✅ File-disjoint from #1154 (act2) and #1157 (R5 act plugins).
- ✅ 2 files, +70/-6: appropriately scoped for audit fix.

No concerns. Fix PR, well-executed.

PR #1175, opened 2026-06-18
docs(fidelity): #1166 β — audit unified plugins vs student deliverables

jsboigeEpita:
## Epic #1166 (β) — Fidelity audit: unified plugins vs student deliverables

Closes #1166. Read-only re-verification of the #35 integration. Deliverable = a committed **fidelity matrix** mapping each student deliverable → its unified absorber → capability-by-capability coverage with `file:line` evidence on both sides, plus per-gap follow-up recommendations and an aggregate verdict.

### Method
Triple grounding (SDDD): student deliverable source ↔ unified plugin source ↔ capability mapping. Student dirs are **SANCTUARY** — this audit never edits them. Evidence cited as `file:line` on both sides.

### Aggregate result
**71 INTEGRATED / 14 PARTIAL / 19 MISSING (8 intentional or N/A)** across 114 mapped capabilities.

| Deliverable | Absorber | Verdict |
|---|---|---|
| `1.4.1-JTMS` | `services/jtms/` | **Preserved** (14/16, near-line-for-line port) |
| `2.1.6 governance` | `agents/core/governance/` | **Preserved+** (17/20, +social-choice layer) |
| `2.3.2 sophismes` | `informal/` + `french_fallacy_adapter` | **Mostly preserved** (6/13) |
| `2.3.3 contre-argument` | `agents/core/counter_argument/` | **Largely preserved** (11/20) |
| `2.3.5 quality` | `agents/core/quality/` | **Preserved+** (11/18, 9 virtues preserved) |
| `2.3.6 local-llm` | `services/local_llm_service.py` | **Partially preserved** (3/8) |
| `1_2_7 dialogique` | `agents/core/debate/` | **Half preserved** (7/11) |
| `3.1.5 Mobile` | `api/mobile_endpoints.py` | **Partially preserved** (2/8) |

**Net**: unification **broadened** capability overall (social-choice, NLI/self-hosted-LLM, SK plugins, capability registry, agentic virtue detectors) but **narrowed** a few specific surfaces. **No silent total-loss of an original capability found.**

### Gaps documented (G1–G9)
| Severity | Gaps |
|---|---|
| **Substantive** | G8 (dialogique dynamic agent layer + `ArgumentationEngine` 10 schemes dropped → LLM-driven adversarial debate) |
[...]

PR #1174, opened 2026-06-18
fix(taxonomy): #1170 T1 — resync fallacies + add virtues taxonomy from Argumentum

jsboigeEpita:
## Track T1 (#1170) — Taxonomy + virtues resync from Argumentum

Closes #1170. Parent Epic #1165 (Culmination). Owner po-2023.

User mandate: *"une taxonomie complète des fallacies, et bien fournie pour les vertus (resync avec le dépôt Argumentum)."*

### Fallacy taxonomy — resynced
`argumentum_fallacies_taxonomy.csv` + `taxonomy_full.csv` brought to the latest Argumentum canonical version:
- **1408 unique PKs** (fixed duplicate PKs **520** + **1000**; **+2 PKs** 1406, 1407).
- Schema adds `ar` / `es` / `zh` / `fa` language columns + `latin1` column.
- Corrected `crossLink_Leverages` typo; ~1014 content refreshes.
- **8 top-level `Famille` values UNCHANGED** (Abus de langage, Argument fallacieux, Erreur de raisonnement, Erreur mathématique, Influence, Insuffisance, Obstruction, Tricherie) — descent + act-narration family enumeration holds. The FR↔EN alias table is schema-compatible (same `Famille`/`path` columns).
- `taxonomy_medium.csv` (28-label flat depth-1+2 index) is **family-consistent** with the resync and left untouched — it's an intentional flat-labels subset, not a source file.

### Virtues taxonomy — added (NEW)
`argumentum_virtues_taxonomy.csv` imported from Argumentum:
- **223 virtues across 8 families** (Argument pertinent, Argument valable, Honnêteté intellectuelle, Langage exact, Présentation intègre, Raisonnement valide, Rigueur mathématique, Échange enrichissant).
- **Force-added**: the `data/` parent dir is gitignored (line 119) and child `!` exemptions are ineffective on Windows `core.ignorecase`. Same precedent as the fallacies CSV (tracked via force-add).
- Imported as the **canonical Argumentum reference**. The quality evaluator's 9 hardcoded virtues (`_evaluator.py VERTUES`) are **NOT refactored** — that would break the FB-38 agentic detectors and their cross-verified measurement (my PR #1129/#1131). The CSV is a traceability artifact, not a runtime dependency.

### DoD — both met
- [x] Project taxonomy + virtues data match the latest Argumentum CSVs (deltas documented above).
- [x] **The descent + quality phases parse the resynced data** — verified via a deterministic 0-LLM/0-JVM smoke test:
[...]

PR #1172, opened 2026-06-18
fix(restitution): #1167 Track D1 — digestion keystone (fallacy attribution + deep_synthesis wiring + Dung/verdict)

jsboigeEpita:
Track **D1** of **Epic #1165** (relaunch #1134). Closes #1167.

## Problem
A direct `spectacular`+`full` run on doc_A (437s, LLM-conducted, healthy) showed **gate=PASS masks broken substance**: rich agentic output was produced then **disconnected** from the report. The diagnostic (grounded, do not re-derive) identified three interconnected losses. This track fixes the architectural heart by **wiring** (never template/invention — anti-pendule #1019/#369).

## D1a — Fallacy attribution (25 orphans → grounded targets)
The writer `_write_hierarchical_fallacy_to_state` was **correct** but received fallacies with empty `family`/`taxonomy_path`/`target` — observed state had 25 entries as `{type, justification}` only. Root cause:
- The wide-net `confirm_fallacy` tool emits no quote/target;
- The per-argument harness attached `source_arg_id` (a real arg_id) but the writer ignored it — it only tried text-match fallbacks on empty fields.

**Fix:**
- `_invoke_hierarchical_fallacy_per_argument` (invoke_callables.py) now enriches each fallacy with a grounded `target_argument` (= the analyzed arg_id) + a verbatim `problematic_quote` span (in-memory only — results are gitignored; privacy HARD).
- The writer resolves `target_argument_id` from `source_arg_id` **first** (when it is a real arg_id in `identified_arguments`), then the text-match fallbacks.

## D1b — deep_synthesis wiring (74s grounded synthesis was dropped)
No `deep_synthesis` key in `CAPABILITY_STATE_WRITERS` → the spectacular phase's grounded FB-18 synthesis + value_gates were never persisted; `state.narrative_synthesis` stayed empty (Acte III reads it, never saw the most global component). Added `_write_deep_synthesis_to_state` (persists `grounded_synthesis` → `state.narrative_synthesis` + value_gates onto `workflow_results`) and registered it. Fail-loud: empty results left as the empty default, never fabricated (#1108/#1019).

## D1c — Dung alignment + verdict band formal credit
[...]

PR #1119, opened 2026-06-15
fix(spectacular): #1115 remove template narrative_synthesis phase (anti-pendule)

jsboigeEpita:
## Summary

Closes #1115. FB-33 (#1113) audit found the `narrative_synthesis` phase was still wired into the spectacular workflow despite FB-31's "always-skipped" claim. Its `build_narrative()` is an f-string over state counts (N arguments, M fallacies...) -> near-identical prose regardless of model -> a **determinization residue (#1109 sec4/5)** that killed richness AND variance. The canonical spectacular deliverable is the LLM-conducted `deep_synthesis` 9-section report (Section 9 fail-loud per FB-31 #1108); the template narrative was only "habillage" surfaced alongside it.

**Fix = subtraction (anti-pendule)**: the LLM path is now the ONLY narrative path on spectacular. No "smarter template" counterweight added.

## Changes

| File | Change | Why |
|------|--------|-----|
| `orchestration/workflows.py` | Remove narrative_synthesis phase from build_spectacular_workflow + drop from deep_synthesis depends_on | The template phase itself (spectacular-only) |
| `visualization/html_report.py` | Guard Narrative section on field presence (Jinja if-block) | Shared renderer: render only when produced; spectacular no longer, standard/full/sherlock still do |
| `tests/.../test_spectacular_workflow_dag.py` | Phase count 17->28, expected set = real 28 phases, +load-bearing test_narrative_template_removed_from_spectacular | Pre-existing DAG debt + #1115 removal guard |
| `tests/.../test_spectacular_regression_suite.py` | TestSpectacularWorkflowGolden expected set + counts to reality (28 phases, 8 levels); replaced fragile exact-completed with load-bearing invariants (no FAILED; non-optional COMPLETED); +removal guard | Fold pre-existing DAG debt (ai-01 R407) |
| `tests/golden/test_spectacular.py` | DoD docstring clarification | Frozen fixture still carries historical narrative; tests validate the frozen artifact, not the live path |

## What is NOT touched

- **DeepSynthesisAgent.ARTIFACT_FIELDS** keeps narrative_synthesis as an OPTIONAL citable field — getattr-defensive (verified), returns None gracefully. No code change needed.
- **standard / full / sherlock_modern workflows** keep the phase (only spectacular is de-castrated per #1115 scope).
- **TestSherlockModernWorkflowGolden** assertions — that workflow legitimately still has the phase. Untouched.
[...]

PR #1176, opened 2026-06-18
chore(synthesis): opportunistic polish — 3 non-blocking Hermes concerns on D1 #1172

jsboigeEpita:
## chore(synthesis): opportunistic polish — 3 non-blocking Hermes concerns on D1 #1172

Addresses the 3 non-blocking review concerns Hermes raised on the merged D1 keystone (PR #1172, commit `b2aadfed`). Cosmetic/diagnostic only — **zero behavior change**, no new guards (anti-pendule).

### Concern 1 — `problematic_quote` field semantics (`invoke_callables.py`)
The field name matches the `IdentifiedFallacy` Pydantic schema, but the value attached by `_enrich_fallacies` is a **match anchor** (the argument's leading text, used by the writer for ID resolution), **not** the exact fallacious phrase. It is never persisted into the narrated report. Added an inline comment at the assignment site so a downstream reader of the intermediate dict isn't misled. **No rename** — renaming would diverge from the schema field name.

### Concern 2 — `except Exception: pass` on value_gates (`state_writers.py`)
Was silently swallowing any state-attr error on the value_gates side-channel persistence. Replaced `pass` with `logger.debug(..., exc_info=True)` so a state-lock regression stays diagnosable. Non-fatal — `narrative_synthesis` (the headline) is set above and is unaffected.

### Concern 3 — double-writer on `state.narrative_synthesis`
`_write_narrative_synthesis` (full path) and `_write_deep_synthesis` (spectacular) both set `state.narrative_synthesis`. They are **workflow-disjoint** (`narrative_synthesis` was removed from spectacular in #1119). Documented the disjointness in the `_write_deep_synthesis` docstring instead of adding an execution guard — the disjointness lives in the workflow phase sets and a guard here would duplicate that contract (anti-pendule).

### Validation
- **mypy strict clean** on both modified files.
- **195 synthesis + restitution tests pass** (no regression).
- Zero behavior change — comment + docstring + debug-log only.

### Scope
Touches `invoke_callables.py` (comment only, ~line 4094) + `state_writers.py` (docstring + 1 debug-log). Disjoint from po-2025's W1 lane (W1 wires dormant reasoners / `_pl_atom` helper; this PR only adds comments to the D1 enrichment block). References #1172 (merged).

Opportunistic — this was flagged by Hermes as "transmis à po-2023 pour polish opportuniste" (R441). Worker idle otherwise (T1 #1174 + β #1175 await admin merge; R1 #1171 gated).

### DIFF
--- a/argumentation_analysis/orchestration/invoke_callables.py
+++ b/argumentation_analysis/orchestration/invoke_callables.py
@@ -4091,6 +4091,11 @@ async def _invoke_hierarchical_fallacy_per_argument(
                     return
                 # Verbatim span: cap to keep the quote a tight anchor (privacy +
                 # match precision — the writer matches on desc[:60] prefixes).
+                # NB: the field name ``problematic_quote`` matches the
+                # IdentifiedFallacy schema, but here it is NOT the exact
+                # fallacious phrase — it is a match anchor (the argument's
+                # leading text) used by the writer for ID resolution only, and
+                # is never persisted into the narrated report.
                 quote_span = arg_text.strip()[:200]
                 for f in fallacies:
                     if not isinstance(f, dict):
--- a/argumentation_analysis/orchestration/state_writers.py
+++ b/argumentation_analysis/orchestration/state_writers.py
@@ -938,6 +938,16 @@ def _write_deep_synthesis_to_state(
         trace for downstream consumers).
     Empty/unavailable results are left as the empty default so the gap is
     reported honestly (fail-loud, #1108/#1019 — never fabricate here).
+
+    Workflow disjointness (double-writer note): both this writer and
+    ``_write_narrative_synthesis_to_state`` (the ``full`` path) set
+    ``state.narrative_synthesis``. They are never active in the same run —
+    ``narrative_synthesis`` was removed from the ``spectacular`` workflow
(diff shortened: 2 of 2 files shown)


jsboigeEpita (commented):
**[Hermes]**

LGTM. Polish diagnostic ciblé, zero behavior change confirmé.

Vérifié :
1. `logger` bien défini (`state_writers.py:13`, `getLogger("UnifiedPipeline")`) — le `logger.debug` dans le bloc except est safe, pas de risque NameError dans le path d'erreur.
2. Concern 1 (sémantique `problematic_quote`) — commentaire précis : distingue l'anchor de match (ID resolution) du verbatim persisté. Clarifie un nom de champ trompeur.
3. Concern 2 (disjointness docstring) — documente le contrat de phase-set non-évident entre `deep_synthesis` (spectacular) et `narrative_synthesis` (full). Justifie l'absence de garde (anti-pendule).
4. Concern 3 (`pass`→`logger.debug`) — préserve le fail-loud contract (la synthèse headline reste set au-dessus) tout en gardant le state-lock diagnosable. Bonne granularité.

No secrets. No regression risk. Ready to merge.

PR #1198, opened 2026-06-20
FP-6 #1197 — FOL invoke-callable fail-loud: stop re-castrating None into a fabricated verdict

jsboigeEpita:
Closes #1197. Sibling of FP-3 #1195. Epic #1191.

## Firsthand constat (the user's "constate par toi-même" mandate)
Re-ran corpus_A on the FP-3-fixed main `72586016`. PL now returns **real PySAT verdicts** (a ~30-atom KB that used to OOM now returns `satisfiable: True` with a model in ms). But the FOL section showed `consistent: False, confidence: 0.4` — while the **live log showed the reasoner OOM'd** and the FP-3 handler honestly returned `degraded (None)`. The honesty was destroyed downstream.

## Root cause — théâtre regrows above the handler (R405)
`_invoke_fol_reasoning` (`invoke_callables.py`) re-castrated the handler's tri-state:
- **main path**: `"consistent": bool(is_consistent)` → `bool(None) == False` fabricated a definite "KB is inconsistent" verdict (`confidence 0.4`).
- **isolation retry**: hardcoded `"consistent": True` after only checking each formula *parses* — never a consistency check on the combined KB.

## Fix (subtraction, #1019 fail-loud)
- Propagate the handler tri-state unchanged: `None` ⇒ unverified (confidence 0.0), `True` ⇒ 0.8, `False` ⇒ real inconsistency verdict (0.4). Mirrors the conversational path which already does `fol_out.get("consistent")`.
- Isolation retry now runs a **real combined consistency check** on the parseable survivors and propagates its tri-state; if that check itself OOMs/raises ⇒ `None` (unverified), never a fabricated `True`.

## Tests (LLM + JVM mocked, no API/no JVM)
- `None` (handler OOM) ⇒ `consistent=None, confidence=0.0` (regression: was `False/0.4`)
- `True` ⇒ `consistent=True, confidence=0.8`
- `False` ⇒ `consistent=False, confidence=0.4` (a real inconsistency, distinct from None)
- isolation branch ⇒ no fabricated `True`; combined re-check tri-state propagates
4 passed.

## Scope note — pre-existing red on main (NOT this PR)
While verifying, I found **5 pre-existing test failures on main `72586016`** (they fail with this PR's changes stashed): obsolete PL tests asserting the *old graceful Python fallback* that FP-3/#1019 deliberately removed (now a fail-loud `RuntimeError: ... JVM/Tweety required`). Tracked separately as FP-7 #1199 — out of scope here to keep this FOL fix focused.

## Privacy
Opaque IDs only. The firsthand report is gitignored local — no raw content in this PR.

🤖 Coordinator ai-01 (self-pickup: surgical fix root-caused firsthand; both workers busy on FP-4/FP-5)

### DIFF
--- a/argumentation_analysis/orchestration/invoke_callables.py
+++ b/argumentation_analysis/orchestration/invoke_callables.py
@@ -5560,20 +5560,24 @@ async def _invoke_fol_reasoning(
             # `None` (unverified), never a fabricated True.
             iso_consistent: Optional[bool] = None
             iso_msg = "combined consistency unverified"
-            try:
-                combined_bs = "\n".join(
-                    str(f) for f in fol_signature + [""] + valid_formulas
-                )
-                iso_consistent, iso_msg = await asyncio.to_thread(
-                    bridge.check_consistency, combined_bs, "first_order"
-                )
-            except Exception as iso_err:  # combined check failed → honest unverified
-                logger.warning(
-                    "FOL isolation: combined consistency check failed (%s) — "
-                    "reporting unverified (None), not fabricated True.",
-                    iso_err,
-                )
-                iso_consistent = None
+            # bridge is non-None whenever valid_formulas is non-empty (the loop
+            # above is guarded by `if bridge is not None`), but narrow explicitly
+            # for the type checker; None ⇒ stay unverified (never fabricate True).
+            if bridge is not None:
+                try:
+                    combined_bs = "\n".join(
+                        str(f) for f in fol_signature + [""] + valid_formulas
+                    )
+                    iso_consistent, iso_msg = await asyncio.to_thread(
+                        bridge.check_consistency, combined_bs, "first_order"
+                    )
+                except Exception as iso_err:  # combined check failed → unverified
+                    logger.warning(
+                        "FOL isolation: combined consistency check failed (%s) — "
+                        "reporting unverified (None), not fabricated True.",
+                        iso_err,
+                    )
+                    iso_consistent = None
             return {
                 "formulas": valid_formulas,
                 "fol_signature": fol_signature,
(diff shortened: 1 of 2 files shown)


jsboigeEpita (commented):
**[Hermes]** COMMENT (self-review cap, po-2026).

**Root cause confirmed.** The `bool(is_consistent)` cast on the main path (`bool(None) == False`) did fabricate a definite "KB inconsistent" verdict (`confidence 0.4`) whenever the FP-3 handler honestly returned `None` (reasoner OOM/degraded). The isolation branch's hardcoded `consistent: True` after a parse-only loop was a second, independent théâtre regrowth (R405). Both are genuine correctness bugs, not cosmetic.

**Fix verification (read the full diff):**
- Main path: `"consistent": is_consistent` propagates the tri-state unchanged; confidence ladder `True→0.8 / None→0.0 / False→0.4` is correct and mirrors the existing conversational path (`fol_out.get("consistent")`).
- Isolation branch now runs a *real combined* `check_consistency` on the parseable survivors and propagates its tri-state; the `except` → `None` (unverified) is the right fail-loud posture. The `combined_bs` join with the signature is a sensible belief-set format.
- `Optional` and `logger` are already imported in the file — no missing-symbol risk.
- Security scan: clean (no tokens/secrets in diff).

**Tests (mocked LLM + JVM, no API/JVM):** the 4 cases pin the regression contract precisely — `None≠False`, `True→0.8`, `False→0.4`, isolation-no-fabricated-True. Assertions match the implementation exactly. 4 passed as reported.

**One observation (non-blocking):** a large fraction of the +87/-36 in `invoke_callables.py` is unrelated black reformatting (multi-line `if`/list-comp wrapping) outside the FOL functions. Harmless and PEP8-conformant, but it inflates the apparent scope of a "surgical subtraction" fix. Fine to merge as-is; worth keeping in mind for focused-diff hygiene on future hotfixes.

LGTM-leaning — root cause identified, fix complete, regression pinned.

PR #1212, opened 2026-06-21
fix(logic): #1205 SPASS modal ctor wiring (1-arg path) + real test (sibling of eprover fix)

jsboigeEpita:
## Summary

Implements the **R458 dispatch** from ai-01 on **#1205** (FP-9, L2): fix the SPASS modal reasoner no-arg constructor bug (sibling of the EProver regression #1202) + add a real non-mocked integration test.

The SPASS modal reasoner had the **exact same bug** EProver had pre-#1202: `_get_spass_reasoner` called `SPASSMlReasoner()` with no argument, but Tweety 1.28+/1.29 only exposes `SPASSMlReasoner(String)` / `SPASSMlReasoner(String, Shell)` (**verified via `javap`** on the bundled jar, both 1.28 and 1.29). The construction error was swallowed by a bare `except` → SPASS was never wired, yet the pipeline reported it as the configured modal solver (formal theater, modal side of the #1196/#1202 regression class).

## Fix (mirror of the EProver consumer in `fol_handler.py`)

- New `_get_spass_path()` module-level helper reads `EXTERNAL_TOOL_PATHS['spass']` (mirrors `_get_eprover_path`).
- `_get_spass_reasoner` reads the registered path and builds `SPASSMlReasoner(JString(path))` (1-arg ctor).
- **Fail-loud** `RuntimeError` if the path is unset (anti-theater #1019): a missing SPASS binary surfaces as an honest "cannot decide", NOT a silent fallback to `SimpleMlReasoner`.

## Real test (`test_spass_real.py`, new)

Skip-if-binary-absent, modeled on #1210's real-EProver test. Runs the REAL SPASS binary through the production path (`TweetyBridge.check_consistency(bs, "K")` → `ModalHandler.is_modal_kb_consistent` → `SPASSMlReasoner(path).isConsistent`). **Skips cleanly on po-2023** (no binary); ai-01 (binary present) runs it firsthand before merge. Anti-théâtre: EXECUTES the binary where present, NOT "green by skipping everywhere".

## Unit tests adapted

The 5 existing SPASS tests in `test_eprover_spass_integration.py` mocked the no-arg ctor; now mock `_get_spass_path` + the 1-arg ctor. Added 2 fail-loud contract tests: (a) binary absent → `RuntimeError("SPASS binary not detected")`; (b) construction error → `RuntimeError`, no silent fallback.

## Verification

- **7/7 `TestModalHandlerSPASS` pass** (5 adapted + 2 new fail-loud contracts).
[...]

PR #1216, opened 2026-06-21
fix(logic): FP-12 #1215 make DL/DeLP/CF2 actually decide

### LINKED ISSUE #1215: FP-12 — DL/DeLP/CF2 wiring fixes: make them DECIDE or honest fail-loud (Epic #1191)
Fix-track from FP-4 diagnosis (#1194 / report `FP4_EXTENSION_EMPTINESS_DIAGNOSIS.md`). Three concrete wiring bugs, each firsthand-confirmed by ai-01 (R460). Same arc as eprover/modal: a reasoner that *looks* wired but never actually decides = théâtre (#1019). Fix each to **truly decide**, or honest fail-loud `None` where the procedure genuinely can't run — never fabricate.

## 1. DL consistency is hardcoded `True` (théâtre) — `dl_handler.py:~163`

```python
result = reasoner.query(kb, self._AtomicConcept("⊤"))   # ⊤ via AtomicConcept ctor is wrong
# An inconsistent KB entails everything
return True, "Knowledge base is consistent."            # <-- result IGNORED, always True
```

`result` is computed then thrown away; the method **unconditionally returns `True`**. So DL consistency never decides — worse than the diagnosis (which only flagged the `⊤`/AtomicConcept cast). **Fix**: use a real consistency procedure (native method if `NaiveDlReasoner` exposes one — verify via `getMethods()` introspection like the modal `isConsistent` finding; else bottom-entailment: inconsistent ⟺ KB entails `⊥`/BottomConcept). Return the **real** bool. Decide on a synthetic consistent KB → True, inconsistent KB → False.

## 2. DeLP feeds raw prose to the parser — `invoke_callables.py:3623`

```python
program_text = context.get("program", input_text[:500])
```

When no structured `program` is in context it defaults to the **first 500 chars of raw prose** → `DelpParser` rejects it (correctness) AND raw text reaches the parser (privacy smell). **Fix (subtraction, anti-pendule)**: remove the `input_text[:500]` default; when no real DeLP program is provided, honest fail-loud (`DeLP unavailable: no program provided`), do NOT fabricate from prose. (Same line 1170 `targets = [input_text[:500]]` — check if it has the same raw-prose smell.)

## 3. CF2 reasoner FQN may be missing in Tweety 1.29 — `af_handler.py:20`

```python
"cf2": "org.tweetyproject.arg.dung.reasoner.CF2Reasoner",
```

Diagnosis reported `CF2Reasoner class not found` at runtime. **Fix**: verify the real FQN via `javap`/JVM introspection against the bundled Tweety jar (cf. #1179). If it exists under a different name → correct it (CF2 decides). If genuinely absent from 1.29 → gate it out with an honest "unavailable in bundled Tweety" message (do NOT silently return empty as if it decided).

## DoD

- [ ] DL `is_consistent` returns a **real** bool (decides synthetic consistent→True / inconsistent→False), result no longer ignored. Real non-mocked test.
- [ ] DeLP raw-prose `input_text[:500]` default removed; honest fail-loud when no program. (Check line 1170 too.)
- [ ] CF2 FQN verified via introspection: corrected (decides) or honestly gated-out (documented). Real non-mocked test for whichever path.
- [ ] Each fix has a real test that DECIDES (not skip-everywhere — these are pure-Java Tweety reasoners, no external binary). green-by-skip = re-théâtre (#1019).

## Privacy HARD
Synthetic atoms only. The DeLP fix specifically REMOVES a raw-text path — verify no `input_text`/`raw_text` reaches any parser or state.

## Anti-pendule
- DL: don't keep returning hardcoded True — use the real verdict (subtract the ignored line).
- DeLP: subtract the raw-prose default, don't add a prose→program translator here (that's depth-track).
- CF2: don't silently swallow the missing class as empty-=-decided; honest-absent.

Epic #1191. Carries forward the FP-4 wiring bucket (b)+(d).

jsboigeEpita:
## FP-12 #1215 — DL / DeLP / CF2: actually decide (bottom-entailment, fail-loud, honest gate-out)

Three reasoners that *seemed* wired but never actually decided — the same
**#1019 silent-theater regression family** as the modal/DL bugs in #1212/#1213.
Each one returned a verdict that was either hardcoded, fed garbage, or promised
a class that does not exist.

### Fix 1 — DL `is_consistent` (fabricated True → real bottom-entailment)

`dl_handler.is_consistent` computed `reasoner.query(kb, TopConcept)` and then
**ignored the result**, unconditionally returning `(True, "consistent")`.
Querying TopConcept cannot distinguish consistent from inconsistent anyway (a
consistent KB entails Top trivially, but so does an inconsistent one which
entails everything). This was worse than "degraded": a fabricated verdict.

`NaiveDlReasoner` exposes only `query` (no `isConsistent` — same gap as the
modal reasoners, #1212). The deciding procedure is **bottom-entailment**: a KB
is consistent iff it does **not** entail BottomConcept. Assert
`(witness: Bottom)` and ask whether the KB entails it.

**Firsthand-verified** against a real JVM (probe + integration test):
- consistent KB `(john:human)` → `query(witness:Bottom)=False` → **True**
- inconsistent KB `(john:human) ∧ (john:¬human)` → `query=True` → **False**

Returns `(None, msg)` on a reasoner exception — fail-loud degraded (#1019),
never a fabricated `False` disguised as "inconsistent". `is_consistent` is now
`Tuple[Optional[bool], str]`.

### Fix 2 — DeLP raw-prose default (parse-garbage + corpus leak → fail-loud)

`_invoke_delp` defaulted `program_text = context.get("program", input_text[:500])`.
DeLP syntax (`fact.`, `head <- body.` strict, `head -< body.` defeasible) is
**not prose** → guaranteed parse failure, **and** a corpus-text leak into the
parser (privacy violation). Fix is **subtraction**: drop the raw-text default;
when no program is in context, return `status="unavailable"` with the DeLP
syntax hint. The raw corpus string never reaches the parser (verified by the
privacy-guard test: a `SECRET_CORPUS_LEAK_*` input is rejected upstream and
appears nowhere in the result).

### Fix 3 — CF2 honest gate-out (false promise → removed)

`CF2Reasoner` is **NOT shipped** in the vendored Tweety build. Introspected
firsthand: the FQN raises `ClassNotFoundException` in both 1.28 and 1.29; no
`SimpleCF2Reasoner` / `prover.CF2Reasoner` / `prover.Prover` variant exists
either. Advertising cf2 as supported while the class is absent was a false
promise. Honest gate-out: cf2 is removed from `SEMANTICS_REASONERS` →
`_get_reasoner("cf2")` now raises `ValueError` (not a silent
`{"error": ...}` dressed as a result). The other 10 semantics are untouched.

---

## Design notes

- **Anti-pendule**: each fix *subtracts* the bug (the hardcoded `return True`,
  the raw-text default, the false-promise dict entry). No counterweight added.
- **Anti-theater #1019**: DL returns `Optional[bool]` — `None` = degraded,
  never a fabricated verdict. DeLP returns `unavailable`, never a parse-on-prose
  masquerading as "analyzed".
- **Privacy HARD**: the DeLP fix *removes* a raw-text path. Verified no
  `input_text` / `raw_text` reaches a parser or the state. Synthetic atoms only
  (`human`/`john`), 0 corpus content.

---

## DoD (from dispatch #1215)

- [x] **DL `is_consistent` → vrai bool** (synthetic consistent→True /
      inconsistent→False), result plus ignoré. **Test réel non-mocké.**
      → `TestDLDecidesViaBottomEntailment` (real `DLHandler` + real JVM):
      `test_consistent_kb_decides_true`, `test_inconsistent_kb_decides_false`.
- [x] **DeLP défaut raw-prose retiré** ; fail-loud si pas de programme.
      → `TestDeLPNoRawTextFailLoud` (3): no-program→`unavailable`; raw-corpus
      string never reaches parser (privacy guard); explicit program forwarded.
- [x] **CF2 FQN vérifié par introspection** : gaté honnêtement (class absente).
      **Test réel pour le chemin retenu.**
      → `TestCF2HonestGateOut` (2): cf2 not in `SEMANTICS_REASONERS` /
      `supported_semantics` (real import); `_get_reasoner("cf2")` raises
      `ValueError` (real `AFHandler` + real JVM).
- [x] **Chaque fix a un test qui DÉCIDE** (pas skip-partout — reasoners pure-Java
      Tweety). The DL + CF2 tests run the **real** `NaiveDlReasoner` /
      `AFHandler` against a real JVM (no skip, no mock). DeLP fail-loud is
      deterministic Python (the guard fires before any JVM call).

green-by-skip = re-théâtre: **non** — 5 of 7 tests exercise real Tweety classes.

---

## Validation

- **69 FP-12 tests pass** (new integration 7 + bumped DL/AF suites).
- **Full logic suite**: 337 passed, 43 skipped, **1 failed**
  (`test_fol_query_dispatches_to_eprover` — **pre-existing**, confirmed via
  `git stash` + run on clean main `1e379fd5` where the same test fails; not a
  regression of this PR).
- **mypy strict** on `invoke_callables.py` (CI-gated): clean.
- **black + flake8**: clean.
- **Spectacular DAG intact** (`test_spectacular_workflow_dag.py`): 16 passed —
  the phases (`description_logic`, `delp_reasoning`, `dung_extensions`) already
  exist; this PR only changes their *internal* behavior, not the workflow
  structure.

Closes #1215

### DIFF
--- a/argumentation_analysis/agents/core/logic/af_handler.py
+++ b/argumentation_analysis/agents/core/logic/af_handler.py
@@ -7,7 +7,12 @@ from .tweety_initializer import TweetyInitializer
 logger = logging.getLogger(__name__)
 
 
-# Mapping from semantics name to Tweety reasoner class path
+# Mapping from semantics name to Tweety reasoner class path.
+# #1215 (FP-12): ``cf2`` removed — ``CF2Reasoner`` is NOT shipped in the vendored
+# Tweety build (introspected firsthand: the FQN raises ClassNotFoundException in
+# both 1.28 and 1.29; no SimpleCF2Reasoner / prover.* variant exists either).
+# Advertising it as supported while the class is absent was a false promise;
+# honest gate-out instead of a silent {"error": ...} dressed as a result.
 SEMANTICS_REASONERS = {
     "preferred": "org.tweetyproject.arg.dung.reasoner.SimplePreferredReasoner",
     "grounded": "org.tweetyproject.arg.dung.reasoner.SimpleGroundedReasoner",
@@ -17,7 +22,6 @@ SEMANTICS_REASONERS = {
     "conflict_free": "org.tweetyproject.arg.dung.reasoner.SimpleConflictFreeReasoner",
     "semi_stable": "org.tweetyproject.arg.dung.reasoner.SimpleSemiStableReasoner",
     "stage": "org.tweetyproject.arg.dung.reasoner.SimpleStageReasoner",
-    "cf2": "org.tweetyproject.arg.dung.reasoner.CF2Reasoner",
     "ideal": "org.tweetyproject.arg.dung.reasoner.SimpleIdealReasoner",
     "naive": "org.tweetyproject.arg.dung.reasoner.SimpleNaiveReasoner",
 }
@@ -27,7 +31,8 @@ class AFHandler:
     """
     Handles Abstract Argumentation Framework (AF) operations using TweetyProject.
     Supports multiple Dung semantics: preferred, grounded, stable, complete,
-    admissible, conflict-free, semi-stable, stage, CF2, ideal, naive.
+    admissible, conflict-free, semi-stable, stage, ideal, naive.
+    (CF2 is NOT shipped in the vendored Tweety build — see SEMANTICS_REASONERS.)
     """
 
     def __init__(self, initializer_instance: TweetyInitializer):
@@ -124,8 +129,8 @@ class AFHandler:
             arguments: A list of argument names.
             attacks: A list of attacks [source, target].
             semantics: The semantics to use. Supported: preferred, grounded, stable,
-                      complete, admissible, conflict_free, semi_stable, stage, cf2,
-                      ideal, naive.
+                      complete, admissible, conflict_free, semi_stable, stage,
+                      ideal, naive. (cf2 is NOT available in the vendored build.)
 
         Returns:
             Dict with keys: semantics, extensions, statistics.
--- a/argumentation_analysis/agents/core/logic/dl_handler.py
+++ b/argumentation_analysis/agents/core/logic/dl_handler.py
@@ -53,6 +53,8 @@ class DLHandler:
             self._RoleAssertion = jpype.JClass(f"{dl_pkg}.RoleAssertion")
             self._DlBeliefSet = jpype.JClass(f"{dl_pkg}.DlBeliefSet")
             self._DlSignature = jpype.JClass(f"{dl_pkg}.DlSignature")
+            # Bottom concept for consistency checking (bottom-entailment, #1215).
+            self._BottomConcept = jpype.JClass(f"{dl_pkg}.BottomConcept")
 
             self._DlParser = jpype.JClass("org.tweetyproject.logics.dl.parser.DlParser")
             self._NaiveDlReasoner = jpype.JClass(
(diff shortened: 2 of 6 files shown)


jsboigeEpita (commented):
**[Hermes, po-2026]** COMMENT (self-review cap) — leaning LGTM, strong PR.

**Scope**: FP-12 #1215 — DL/DeLP/CF2 reasoners that *seemed* wired but never actually decided (#1019 theater family). +250/-31 across `af_handler.py`, `dl_handler.py`, `invoke_callables.py` + new integration test (155 LOC).

**Deep-read findings**:
1. **DL `is_consistent` bottom-entailment** ✅ — the old code ran `reasoner.query(kb, ⊤)` then *ignored the result* and returned hardcoded `(True, "consistent")`. Querying ⊤ cannot distinguish consistent/inconsistent anyway (inconsistent KBs entail everything). New impl asserts `(witness: ⊥)` and queries entailment of bottom — correct deciding procedure for `NaiveDlReasoner` which exposes only `query` (no `isConsistent`). Exception path now returns `(None, msg)` fail-loud instead of fabricated `False`. Solid.
2. **DeLP raw-text → parser** ✅ — `_invoke_delp` no longer defaults `program_text` to `input_text[:500]`. DeLP syntax ≠ prose, so the old default guaranteed a parse failure *and* leaked corpus text into the parser (privacy). Honest `unavailable` + clear syntax hint. Good subtraction fix.
3. **CF2 gate-out** ✅ — `CF2Reasoner` removed from `SEMANTICS_REASONERS`. Docstrings/`_invoke_dung_extensions` updated to drop the false "all 11 semantics" claim. `_get_reasoner("cf2")` now raises `ValueError` instead of silently producing an empty result dressed as "decided". Correct given the class isn't in the vendored Tweety build.

**Tests**: `test_fp12_dl_delp_cf2.py` is REAL (JVM fixture, no mocks). Consistent/inconsistent KB pair asserts the verdict actually flips; DeLP no-program path asserts `unavailable`; privacy guard uses `"SECRET_CORPUS_LEAK_"` synthetic atom and asserts it never appears in the result. Synthetic atoms only (human/john) — 0 corpus content. These are the right tests.

**Concern (non-blocking)**: `test_raw_text_never_reaches_parser` and the `test_no_program` methods are declared `async def` but I don't see a pytest-asyncio marker/fixture in the diff slice — confirm the test module has `@pytest.mark.asyncio` or an asyncio mode configured, else these will be silently skipped/collected as sync-warnings. Worth a `pytest --collect-only` pass.

Verdict: exemplary "honest fail-loud, no theater" fix. Merge after confirming async tests actually execute.

PR #1221, opened 2026-06-21
fix(logic): #1219 modal pipeline reaches SimpleMlReasoner (no more valid=None)

jsboigeEpita:
## Summary

Closes #1219 — the modal logic capability (fixed in unit tests by #1212/#1214)
was **never reached by the spectacular pipeline**: every `modal` phase returned
`valid=None` because `_invoke_modal_logic` could not load a solver. Surfaced by
FP-13 #1218's depth-parity matrix (modal cell = honest-empty, the théâtre
suspect the DoD required flagging).

## Root cause (two bugs, both masked by the SPASS cascade)

`_invoke_modal_logic` (the spectacular `modal` phase) had **two** construction
bugs that unit tests could not catch (they patch `asyncio.to_thread`, so the
real `ModalHandler` never runs):

1. **Unconditional SPASS force-set** — `object.__setattr__(settings,
   "modal_solver", ModalSolverChoice.SPASS)` ran on *every* call. With the SPASS
   binary absent, `_get_active_reasoner` raised `RuntimeError` *before*
   `is_modal_kb_consistent` could run → fell to the TweetyBridge fallback.
2. **Bare `TweetyInitializer()`** — does NOT initialize the modal parser/reasoner
   (class attrs start `None`; set only by `initialize_modal_components`, which
   `TweetyBridge.__init__` calls but a bare initializer does not). So
   `ModalHandler.__init__` raised "Modal Parser not initialized" → fell through.

Net: the #1212 `SimpleMlReasoner` query-based consistency fix (proven in
`test_fp11_modal_kb_roundtrip`) was unreachable in the pipeline. Modal always
landed on the no-solver path (`valid=None`).

## Fix (anti-pendule — subtraction of each bug)

- **Remove** the unconditional SPASS force-set + `if modal_solver == "spass"`
  gate. Use the **configured** solver (default `settings.modal_solver = TWEETY`
  → `SimpleMlReasoner`, pure Java, no external binary). SPASS remains opt-in via
  `settings.modal_solver = SPASS` (honored by `_get_active_reasoner`).
- **Add** `initializer.initialize_modal_components()` so `ModalHandler`
  constructs with a real modal parser/reasoner.
- `valid` is `Optional[bool]` (None = undecidable), never fabricated `False`
  (anti-théâtre #1019). The honest `unavailable` fallback (#961) is preserved.

## Verify-the-verification (FP-13 lesson — trust the runtime, not the patched tests)

The patched unit tests passed both before AND after the fix (they never run the
real reasoner). A real end-to-end call against a live JVM confirms the fix:

```
consistent KB   → valid=True   solver='tweety'  "Knowledge base is consistent (tweety)."
inconsistent KB → valid=False  solver='tweety'  "Knowledge base is inconsistent (tweety)."
```

Before the fix, both were `valid=None, solver='unavailable'`.

## Files

- `argumentation_analysis/orchestration/invoke_callables.py` — `_invoke_modal_logic`
  rewrite (remove force-SPASS + gate; add `initialize_modal_components`;
  `Optional[bool]` valid). **mypy strict clean** (CI-gated file).
- `tests/integration/.../test_invoke_modal_logic_reaches_solver.py` — **new**
  non-mocked regression test (skip-if-JVM-absent). Asserts the REAL pipeline
  reaches `SimpleMlReasoner` and decides. Fails if either bug returns.
- `tests/unit/.../test_unified_pipeline.py` — `test_invoke_modal_logic_error`
  was **pre-existing failing** (verified via stash): it asserted the theater
  shape `valid is False` + `"error"`, which #1097's hardening removed. Aligned
  to the honest `valid is None` contract (same staleness-fix pattern as #1217).

## Validation

- `mypy --strict invoke_callables.py` → Success: no issues.
- 25 modal unit tests PASS (incl. load-bearing `test_value_gates` #961/#1097).
- 2 new integration tests PASS (real JVM, real verdicts).
- black + flake8 clean.

## Scope / out-of-scope

This makes modal **decide** in the pipeline via the existing `SimpleMlReasoner`
(not an architectural change). Vendoring a real SPASS CLI build (for more
powerful modal reasoning) remains a separate, optional follow-up — not required
for the depth-parity Epic #1191.

### DIFF
--- a/argumentation_analysis/orchestration/invoke_callables.py
+++ b/argumentation_analysis/orchestration/invoke_callables.py
@@ -5710,10 +5710,34 @@ async def _invoke_nl_to_logic(
 async def _invoke_modal_logic(
     input_text: str, context: Dict[str, Any]
 ) -> Dict[str, Any]:
-    """Invoke modal logic analysis via TweetyBridge or external SPASS solver.
+    """Invoke modal logic consistency via the configured Tweety reasoner.
 
-    Routes to SPASS when context["modal_solver"] == "spass" (#479).
-    Falls back to TweetyBridge or pure-Python heuristic.
+    Primary path: ``ModalHandler.is_modal_kb_consistent`` — the #1212
+    query-based consistency probe (the KB is inconsistent iff it entails a
+    ground contradiction), run with the **configured** modal solver. The
+    default ``settings.modal_solver`` is ``TWEETY`` → ``SimpleMlReasoner``
+    (pure Java, decides without any external binary — #1205, firsthand-
+    verified). SPASS remains opt-in via ``settings.modal_solver = SPASS``
+    (honored by ``ModalHandler._get_active_reasoner``), useful once a real
+    SPASS CLI build is vendored.
+
+    #1219 (anti-pendule): the previous implementation force-set
+    ``settings.modal_solver = SPASS`` *unconditionally* on every call. With
+    the SPASS binary absent (the current state), ``_get_active_reasoner``
+    raised ``RuntimeError`` *before* ``is_modal_kb_consistent`` could run,
+    so the #1212 ``SimpleMlReasoner`` fix — proven in unit tests
+    (``test_fp11_modal_kb_roundtrip``) — was **never reached in the
+    pipeline**: every spectacular ``modal`` phase fell through to the
+    no-solver path (``valid=None``). The force-set was the bug; subtracting
+    it lets the configured ``TWEETY`` default decide. ``ModalSolverChoice``
+    is no longer imported here (no per-call override → config is the single
+    source of truth, mirroring ``ModalHandler``'s own design).
+
+    ``valid`` is ``Optional[bool]``: a real ``True``/``False`` when the
+    solver decided, ``None`` when it could not (parse error, undecidable
+    signature) — honest, never fabricated (#1019). Falls back to
+    ``TweetyBridge.execute_modal_query`` then to an honest ``unavailable``
+    report (#961) — no heuristic that fabricates a verdict.
     """
     formulas = context.get("formulas", [input_text])
     if not isinstance(formulas, list):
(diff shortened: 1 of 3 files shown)


jsboigeEpita (commented):
**[Hermes, po-2026]** COMMENT (self-review cap) — `fix(logic): #1219 modal pipeline reaches SimpleMlReasoner`. +179/-44, 3 files (1 rewrite + 1 new test + 1 unit test update).

**Root-cause verified (incident checklist):** two masked construction bugs, both correctly diagnosed.
1. **Unconditional SPASS force-set** — `object.__setattr__(settings, "modal_solver", SPASS)` ran on *every* call. With the SPASS binary absent, `_get_active_reasoner` raised before `is_modal_kb_consistent` could run → `valid=None`. The fix removes the force-set + the `if modal_solver == "spass"` gate, letting the configured TWEETY default (SimpleMlReasoner, pure Java) decide. Correct subtraction.
2. **Bare `TweetyInitializer()`** — modal parser/reasoner are class attrs starting `None`, set only by `initialize_modal_components`. The added `initializer.initialize_modal_components()` call fixes this.

**Checklist:**
- ✅ **Security scan**: CLEAN — 0 match `HF_TOKEN|API_KEY|SECRET|PASSWORD|TOKEN=|exec|eval`.
- ✅ **Fix complete**: both bugs subtracted; SPASS remains opt-in via config. `ModalSolverChoice` import dropped (single source of truth).
- ✅ **Anti-regression test**: new `test_invoke_modal_logic_reaches_solver.py` runs the REAL `_invoke_modal_logic` (live JVM, non-mocked) — asserts `valid=True`/`False` with `solver=="tweety"`. Closes the "tests pass but path never executes" hole (FP-13 lesson). Synthetic atoms only.
- ✅ **Honest `Optional[bool]`**: `valid` is `is_consistent` (not `bool(...)`), None propagates when undecidable — anti-théâtre #1019. Updated `test_invoke_modal_logic_error` asserts `valid is None` (not old theater `False`+`error`).
- ✅ **async tests**: no explicit `@pytest.mark.asyncio` — consistent with repo convention (asyncio_mode=auto). Not a concern.
- ✅ **Fallback chain preserved**: ModalHandler → TweetyBridge → honest unavailable.

**One non-blocking note:** the integration test SKIPs where JVM can't start (conftest autouse). Worth confirming it runs GREEN on at least one CI runner with a JVM — skip-everywhere would defeat the guard (the PR's own anti-théâtre framing flags this).

**Verdict:** solid, well-documented fix. Root cause complete, anti-regression test genuine. CI-JVM availability is the only pre-merge confirmation.

PR #1223, opened 2026-06-21
fix(fp5): FP-14 #1222 refresh depth-parity matrix post-#1221; modal solver gap resolved

jsboigeEpita:
## Summary

Closes #1222 (FP-14) — refreshes the FP-5 formal-richness depth-parity matrix at
the post-#1221 base (`883fd770`) and confirms the #1219 modal-pipeline-solver
gap is **resolved**: the spectacular `modal` phase now reaches
`SimpleMlReasoner`. The refresh surfaced a **new, distinct gap**
(NL→modal-KB translation) and a **classifier over-label**, both addressed here.

## #1219 — RESOLVED at the pipeline level

`_invoke_modal_logic` (PR #1221, merged `883fd770`) no longer force-sets `SPASS`
and now calls `initialize_modal_components()`. The A/B/C re-run confirms the
solver is reached:

| | pre-#1221 (FP-13) | post-#1221 (FP-14) |
| --- | --- | --- |
| `solver` evidence | `null` (`"unavailable"`) | **`"tweety"`** (SimpleMlReasoner) |
| run log | no modal parser invocation | `MlParser.parseBeliefBase` runs (raises `ParserException`) |
| `valid` | `None` (solver never ran) | `None` (solver ran, **KB malformed**) |

Verified two ways (FP-13 lesson — trust the runtime, not the classifier): the
matrix cell `verdict:"tweety"` AND the run log where `MlParser.parseBeliefBase`
raises — pre-#1221 that parser was never invoked.

## Modal cell — `empty (no solver)` → `degraded`

The solver is reached, but `valid=None` because the corpus→modal-KB translation
produces malformed grammar (the parser rejects it):

- doc_A: `Predicate '<url-fragment>' has not been declared` (URL fragment as predicate).
- doc_C: `Missing '=' in sort declaration '<prose-greeting-fragment>'`.

`is_modal_kb_consistent` catches the parse failure and honestly returns
`(None, "Modal KB parse error (consistency undetermined, tweety)")`. This is
**neither a clean `real-verdict` nor `honest-absent`** (the DoD's binary): the
corpus has modal-ish content, but the NL→modal-KB translation can't render it as
valid `MlParser` grammar. The synthetic unit-test KBs (`type(rain)`,
`[](rain => wet)`) parse and decide cleanly; real political-corpus extraction
does not. **New follow-up** (NL→modal-logic translation quality) — out of scope here.

## Classifier over-label corrected (this PR)

The FP-13 classifier labeled this modal output `real-verdict`:
[...]

Issue #1222, opened 2026-06-21
FP-14 — Refresh depth-parity matrix post #1221 (confirm modal pipeline empty→real-verdict) (Epic #1191)


jsboigeEpita:
Closing measurement for the Epic #1191 depth-parity DoD. The FP-13 matrix (#1220, base `0a5a7a00`) honestly classed the **modal** cell as `empty` because the pipeline `_invoke_modal_logic` never reached a solver (`valid=None`) — flagged as #1219. **#1221 (merged main `883fd770`) fixed that**: modal now decides via `SimpleMlReasoner` in the real pipeline (firsthand-verified `valid=True/False, solver='tweety'`). So the published matrix modal cell is now stale.

## Scope (heavy run — po-2025 lane)

1. **Re-run** `scripts/run_fp5_formal_matrix.py` on corpora A/B/C at base `883fd770`+ (post-#1221).
2. **Confirm the modal cell transition**: modal should now class **real-verdict** where the corpus has modal content, OR **honest-absent** if the corpus genuinely has no modal operators (political prose may legitimately lack `[]`/`<>` content — that is the correct honest class, NOT a regression). Either is acceptable; what is NOT acceptable is `empty`-due-to-pipeline-gap (that bug is fixed) or a fabricated verdict.
3. **Update** `docs/reports/FP5_FORMAL_RICHNESS_MATRIX.md`: mark the #1219 flag **resolved**, refresh the modal row + delta narrative (empty→real-verdict/honest-absent), keep the `modal_fabricated_true=False` anti-théâtre assertion.
4. **Anti-théâtre re-check**: confirm no axis regressed to a fabricated verdict after the modal wiring change (FOL/DL/PL unchanged).

## DoD

- [ ] Matrix re-run A/B/C at post-#1221 base; modal cell no longer `empty`-due-to-pipeline-gap.
- [ ] Modal classed real-verdict (corpus has modal content) or honest-absent (corpus lacks it) — with the evidence line distinguishing the two.
- [ ] `modal_fabricated_true` / `dl_fabricated_true` / `fol_fabricated_true` all `False` (no fabrication introduced).
- [ ] Report updated, #1219 flag marked resolved, delta narrative.

## Privacy HARD
Counts/classifications/aggregates only in report + PR + dashboard. Opaque IDs (corpus_A/B/C). Downstream per-corpus output gitignored (`evaluation/results/`). No raw_text in the report.

## Anti-pendule
[...]

PR #1214, opened 2026-06-21
fix(logic): FP-11 #1213 make modal KB construction parseable by MlParser

jsboigeEpita:
Closes #1213

## Root cause (verified firsthand)

`ModalLogicAgent._construct_modal_kb_from_json` emitted ``constant <name>`` declarations — a **propositional-modal grammar this ``MlParser`` build does NOT implement**. Every modal KB failed to parse:

```
ParserException: Missing '=' in sort declaration 'constant rain'
```

So consistency was **never actually decided**: the old handler swallowed the ``JException`` and returned ``False`` ("inconsistent") — modal theatre, the same regression class as #1019/#1196/#1202. #1212 later surfaced the parse failure as an honest ``None`` (degraded), but the verdict was still never *decided* — the modal axis was structurally incapable of producing a real verdict.

Probe (po-2025, pre-fix, ``TweetyBridge.check_consistency``):
```
constant rain  ->  (None, "...ParserException: Missing '=' in sort declaration 'constant rain'")
```

## Fix — anti-pendule (replace the wrong grammar, not add a parallel path)

``MlParser`` is FOL-modal: propositions are **0-ary predicates declared with ``type(prop)``**; modal operators are ``[]``/``<>``. Firsthand-verified (po-2025, ``SimpleMlReasoner`` pure-Java):

```
type(rain) + type(wet) + [](rain => wet) + rain  ->  (True,  "Knowledge base is consistent (tweety).")
type(rain) + rain + !rain                        ->  (False, "Knowledge base is inconsistent (tweety).")
```

The wrong ``constant X`` grammar is **REPLACED** by ``type(prop)`` across **all 4 emission sites** (grep-confirmed no active ``constant`` remains — only explanatory comments):

1. `_construct_modal_kb_from_json` — the main constructor (the dispatch target).
2. `_parse_modal_belief_set_content` — the round-trip parser (recognises ``type(prop)`` declarations now).
3. `validate_argument` — temporary belief-set construction (same bug).
4. The 4 system prompts + ``TWEETY_MODAL_BNF`` — LLM grammar guidance (the LLM was being told to emit the unparseable form).

`_validate_modal_kb_json` stays coherent: it validates the declared-vs-used JSON contract, independent of the declaration form.

## DoD (#1213)

| Requirement | Status |
|---|---|
| `_construct_modal_kb_from_json` emits ``type(prop)`` instead of ``constant prop`` | ✅ all 4 sites |
[...]

PR #1225, opened 2026-06-22
fix(logic): #1224 build modal KB from nl_to_logic translations, not raw corpus

jsboigeEpita:
## Summary

Closes #1224. The modal phase built its KB from the **raw corpus** instead of the
`nl_to_logic` translations, so `MlParser` choked on URL fragments / prose and
`is_modal_kb_consistent` returned `valid=None` (degraded) on every corpus. This
PR makes `_invoke_modal_logic` read `phase_nl_to_logic_output` (as PL/FOL already
do), sanitize, and declare `type(prop)` so the modal parser accepts the KB —
modal now **decides** (`valid=True`/`False`) on real-pipeline input.

## Root cause (confirmed read-only, then fixed)

`_invoke_modal_logic` read `context.get("formulas", [input_text])`
([invoke_callables.py:5742](argumentation_analysis/orchestration/invoke_callables.py#L5742)).
But the spectacular DAG executor never populates `context["formulas"]`:
`_store_phase_result` ([workflow_dsl.py:651-653](argumentation_analysis/orchestration/workflow_dsl.py#L651-L653))
stores only `phase_*_output`/`_result` and does not merge output keys. And
`_invoke_nl_to_logic` returns `translations` (no `formulas` key). So the fallback
`[input_text]` = the **raw corpus paragraph** → `MlParser.parseBeliefBase` raised
`ParserException` (URL fragment `<url-fragment>` as predicate; prose
`<proper-noun-fragment>` / `<prose-greeting-fragment>` as sort declarations).

PL and FOL avoid this — they read `phase_nl_to_logic_output`
([5246](argumentation_analysis/orchestration/invoke_callables.py#L5246)/[4845](argumentation_analysis/orchestration/invoke_callables.py#L4845)),
extract `[t.formula for t in valid]`, and sanitize. **Modal was the only logic
phase taking the raw-corpus fallback.**

## Fix (anti-pendule: mirror PL/FOL, no new infra)

When `phase_nl_to_logic_output` carries valid translations:
1. Extract `[t["formula"] for t in translations if is_valid]`.
2. Sanitize to clean propositional atoms (`PLFormulaSanitizer`, #537).
3. Declare `type(<atom>)` per unique atom (FP-11 #1214 MlParser contract).
4. Join → modal KB → `is_modal_kb_consistent` decides.

The **direct `{"formulas": ...}` path (hand-written, pre-typed KBs) is unchanged**
— existing `type(...)` declarations are not duplicated (MlParser rejects dupes).
Fail-loud: if the KB still doesn't parse, `(None, "parse error")` → `degraded`,
never a fabricated verdict (#1019).

## Verification (verify-the-verification — real runtime, not classifier)

| Evidence | Result |
| --- | --- |
| Integration `test_nl_translations_consistent_decides_true` (real JVM) | `valid=True`, solver=`tweety` ✅ |
| Integration `test_nl_translations_inconsistent_decides_false` (real JVM) | `valid=False` ✅ |
| Unit: KB source follows translations, not raw corpus | ✅ |
| Unit: invalid-only translations fall through honestly (no fabricated True) | ✅ |
| Unit: direct `{"formulas":...}` path unchanged (regression guard) | ✅ |
| Existing `#1219` hand-written-KB integration tests | 4/4 ✅ |
| Existing modal unit tests (`-k modal`, 3 files) | 14 passed ✅ |
| `mypy --strict invoke_callables.py` | clean ✅ |

This moves the modal depth-parity track forward: the cell should transition
`degraded` → `real-verdict` on a re-run FP-5 matrix (out of scope here; tracked
in #1222/#1223). It is independent of the FP-14 classifier fix (#1223) —
different files, clean rebase on `883fd770`.

### DIFF
--- a/argumentation_analysis/orchestration/invoke_callables.py
+++ b/argumentation_analysis/orchestration/invoke_callables.py
@@ -5739,9 +5739,78 @@ async def _invoke_modal_logic(
     ``TweetyBridge.execute_modal_query`` then to an honest ``unavailable``
     report (#961) — no heuristic that fabricates a verdict.
     """
-    formulas = context.get("formulas", [input_text])
-    if not isinstance(formulas, list):
-        formulas = [str(formulas)]
+    # #1224: build the modal KB from the nl_to_logic translations (as PL/FOL
+    # do), NOT from the raw corpus. The spectacular DAG executor never populates
+    # ``context["formulas"]`` (``_store_phase_result`` sets only
+    # ``phase_*_output``/``_result`` and does not merge output keys), so the
+    # prior ``context.get("formulas", [input_text])`` fallback fed ``MlParser``
+    # the raw corpus paragraph → ``ParserException`` on URL fragments /
+    # prose-as-sort-declarations → ``valid=None`` (degraded) on every corpus.
+    # When nl_to_logic produced formulas, mirror PL: keep the valid ones,
+    # sanitize them to clean propositional atoms (``PLFormulaSanitizer``,
+    # #537), and declare ``type(<atom>)`` per FP-11 #1214 so the modal parser
+    # (an extension of FOL) accepts the KB. The hand-written-KB / direct
+    # ``formulas`` path (tests, pre-typed KBs) stays as-is — never re-declare
+    # atoms an existing KB already declared (MlParser rejects duplicates).
+    nl_out = context.get("phase_nl_to_logic_output") or {}
+    nl_translations = (
+        nl_out.get("translations", []) if isinstance(nl_out, dict) else []
+    )
+    nl_formulas = [
+        str(t["formula"])
+        for t in nl_translations
+        if isinstance(t, dict) and t.get("is_valid") and t.get("formula")
+    ]
+
+    if nl_formulas:
+        # Real-pipeline path (#1224): the modal KB comes from nl_to_logic.
+        # Sanitize to clean propositional atoms, then declare type(prop) per
+        # atom so MlParser accepts the KB. Fail-loud: if the sanitized KB still
+        # does not parse, ``is_modal_kb_consistent`` honestly returns
+        # ``(None, "parse error")`` → degraded, never a fabricated verdict
+        # (#1019). nl_to_logic formulas never carry ``type(...)`` so there is
+        # no duplicate-declaration risk here.
+        formulas = nl_formulas
+        kb_formulas: List[str] = list(nl_formulas)
+        try:
+            from argumentation_analysis.agents.core.logic.pl_formula_sanitizer import (
+                PLFormulaSanitizer,
+            )
+
+            sanitizer = PLFormulaSanitizer()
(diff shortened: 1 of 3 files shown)


jsboigeEpita (commented):
**[NanoClaw]**

Traced the full path: `_store_phase_result` (workflow_dsl.py:638-663) confirms the root cause — it stores only `phase_{name}_result`/`phase_{name}_output`, never `context['formulas']`, so the old `context.get('formulas', [input_text])` fallback fed the modal parser the raw corpus on every spectacular run. The fix correctly mirrors PL/FOL (invoke_callables.py:5246/4845): extract `t['formula']` from valid translations, sanitize, declare `type(<atom>)` per FP-11 #1214. Integration tests prove real decisions on a real JVM (`valid=True`/`False`, solver=`tweety`), not mocked theater. Direct-formulas path preserved verbatim — no duplicate `type(...)` declarations. Fail-loud preserved: sanitizer failure falls to raw formulas → parse fails → honest `(None, 'parse error')`. No fabricated verdict. Security scan clean.

Three concerns, all low severity:

1. **No `logic_type` guard on extraction (invoke_callables.py:5758-5762).** `nl_formulas` pulls every valid translation regardless of type. `PLFormulaSanitizer` is propositional-only — `_SPECIAL_TOKENS` (pl_formula_sanitizer.py:27) has no quantifier/modal tokens, and `_normalize_operators` (line 175) mangles FOL syntax like `man(x)` via `[^a-zA-Z0-9_]→_`. If `phase_nl_to_logic_output` ever carries FOL formulas, sanitization distorts them silently. Mitigated today by `nl_to_logic` defaulting to `logic_type='propositional'` (nl_to_logic.py:197) and honest `valid=None` on parse failure — so it degrades rather than fabricates — but an explicit `if t.get('logic_type') == 'propositional'` filter (as PL does at line 4853) would make the contract self-documenting.

2. **`result['formulas']` semantic mismatch (invoke_callables.py:5762 vs 5853).** In the nl-formulas branch, the returned `formulas` field is the unsanitized `nl_formulas`, while the KB actually evaluated (`belief_set_str`) is the sanitized version + `type(...)` declarations. A consumer reading `result['formulas']` sees different strings than what the solver consumed. Doesn't affect verdict correctness but could confuse downstream tracing.

3. **No modal reasoning on this path (invoke_callables.py:5818-5825).** `nl_to_logic` produces only PL/FOL (nl_to_logic.py:96), so the modality-detection loop will always yield `['none_detected']` and the modal parser evaluates a PL KB. Honest — the PR body states modal depth-parity is tracked in #1222/#1223 — but worth stating explicitly so no one reads 'modal decides True' as 'modal reasoning occurred'. It's PL consistency routed through the modal parser.

COMMENT_WITH_CONCERNS. The fix is correct and well-tested for its stated scope; the concerns are about latent edge cases and documentation, not correctness of the PL path the tests exercise.

PR #1230, opened 2026-06-22
fix(logic): #1227 normalize modal predicate names to MlParser grammar (FP-15 follow-up)

jsboigeEpita:
## Context

Modal-track next notch, surfaced by **FP-15 #1226** (formal-richness matrix re-run post-#1225). After #1225 made `_invoke_modal_logic` build the modal KB from `nl_to_logic` translations (fixing the FP-14 raw-corpus parse failures), the modal cell stayed `degraded` (`valid=None`, `verdict="tweety"`) on all 3 corpora for a **new, distinct cause**: `MlParser` rejects the **compound predicate names** real extraction emits.

## Root cause

`MlParser`'s predicate grammar is `[a-zA-Z][a-zA-Z0-9]*` — **no underscores** — stricter than Tweety PL's `[a-zA-Z_][a-zA-Z0-9_]*` (which `PLFormulaSanitizer` targets). Real `nl_to_logic` extraction produces compound atoms with underscores (e.g. `heavy_rain`). The sanitizer only symbolizes atoms it deems NL-like (>30 chars / punctuation / accents), so a short compound atom passes through underscored, and `type(heavy_rain)` raises `ParserException: Illegal characters in predicate definition`.

## Fix (anti-pendule — subtract the illegal chars)

In the nl-path of `_invoke_modal_logic`, map every MlParser-illegal atom to a fresh legal symbol (`mp1, mp2, ...`), applied **consistently** to both the `type(...)` declarations and the formula bodies so the KB stays sound (no two distinct atoms collapse to one symbol → no spurious consistency). Already-legal atoms are kept verbatim (minimal change). The direct-formulas path (hand-written KBs, tests) is untouched.

This is option 2 (symbolize) from the issue, not option 1 (strip underscores) — strip risks collisions; symbolize with a name→symbol map removes that risk, mirroring PL/FOL.

## Verify-the-verification (FP-13 lesson — runtime, not classifier)

Live-JVM probe confirmed **both** the bug and the fix:

| KB (synthetic compound atoms) | pre-fix | post-fix |
|---|---|---|
| `heavy_rain => wet_ground, heavy_rain` (consistent) | `valid=None` (ParserException) | `valid=True` solver=tweety |
| `heavy_rain, !heavy_rain` (inconsistent) | `valid=None` (ParserException) | `valid=False` solver=tweety |

Two new integration tests guard compound-atom KBs — the #1225 tests used simple atoms (`rain`/`wet`) and **missed** this gap. The inconsistent-compound test specifically proves the symbol map is applied consistently (the atom and its negation map to the same `mpN`, preserving the contradiction).

## Validation

- `mypy --strict argumentation_analysis/orchestration/invoke_callables.py` → **clean** (CI gate).
- **6/6** modal integration tests (4 existing #1219/#1224 + 2 new #1227) pass.
- **3** unit nl-kb + **5** fp11 roundtrip tests pass — no regression (`test_direct_formulas_path_unchanged` confirms the direct path is untouched).
- `black` clean.

## Privacy HARD

Synthetic atoms only (`heavy_rain`/`wet_ground`), 0 corpus token. Diff scanned for entity leaks (lesson R463 — prose ParserException can leak corpus): clean.

## Modal track

`#1219` solver gap → `#1224`/`#1225` KB-source → FP-15 #1226 (measurement, surfaced this) → **#1227 predicate-name normalization** (this PR). The modal cell should move `degraded` → `real-verdict` on the next matrix re-run.

Closes #1227.

### DIFF
--- a/argumentation_analysis/orchestration/invoke_callables.py
+++ b/argumentation_analysis/orchestration/invoke_callables.py
@@ -5753,9 +5753,7 @@ async def _invoke_modal_logic(
     # ``formulas`` path (tests, pre-typed KBs) stays as-is — never re-declare
     # atoms an existing KB already declared (MlParser rejects duplicates).
     nl_out = context.get("phase_nl_to_logic_output") or {}
-    nl_translations = (
-        nl_out.get("translations", []) if isinstance(nl_out, dict) else []
-    )
+    nl_translations = nl_out.get("translations", []) if isinstance(nl_out, dict) else []
     nl_formulas = [
         str(t["formula"])
         for t in nl_translations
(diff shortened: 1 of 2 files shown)


jsboigeEpita (commented):
**[Hermes, po-2026]** — COMMENT (self-review cap)

Solid root-cause fix for #1227. MlParser predicate grammar `[a-zA-Z][a-zA-Z0-9]*` is stricter than Tweety PL (no underscores); `nl_to_logic` emits compound atoms like `heavy_rain` that leaked through `PLFormulaSanitizer` (only symbolizes NL-like atoms >30 chars) and crashed `type(...)` declarations.

**What I verified:**
- The `_legal_symbol()` memoized map substitutes illegal atoms to fresh `mpN` identifiers **consistently** across both `kb_formulas` bodies AND the `_seen_atoms` declarations, so the KB stays sound (no two distinct atoms collapse).
- `_legal_atoms` reservation prevents `mpN` collisions with pre-existing legal atoms named `mp1` etc. Good edge-case handling.
- Keywords returned verbatim; legal atoms untouched -> minimal blast radius.

**Tests** - both new cases are well-targeted:
- `compound_atoms_decide_true`: consistent KB with `heavy_rain` must reach solver and decide True (guards the regression the simple-atom #1225 tests missed).
- `compound_atoms_decide_false`: inconsistent KB - proves the symbol map is consistent (if `heavy_rain` and its negation mapped to different symbols the contradiction would be lost -> spurious True). Key soundness guard.

Security: clean.

**Verdict: LGTM.** Root cause correctly identified, fix is minimal and anti-pendule, tests cover both the regression and the consistency invariant.

PR #1252, opened 2026-06-24
fix(logic): FP-22 #1249 kill _python_dung_fallback theatre — fail-loud (#1019)

jsboigeEpita:
## Summary

FP-22 of Epic #1191 (formal depth-parity) — DoD#3.

Audit of all 5 named `_fallback` functions in `invoke_callables.py`. 4/5 already raise RuntimeError (RA-8 #1053). Only `_python_dung_fallback` was **theatre**: pure-Python combinatorial enumeration of extension sets without any genuine Tweety/JVM call — anti-theater violation (#1019).

**Fix:**
- Both call sites (`TimeoutError` + `afhandler unavailable`) return honest-absent `{"degraded": True, "extensions": {}, "semantics": "unavailable", "note": "..."}` directly
- `_python_dung_fallback` converted to fail-loud stub (raises RuntimeError with actionable message)
- 6 old tests that expected computation replaced with 2 that expect RuntimeError
- Sentinel test confirms `_invoke_dung_extensions` returns degraded dict when JVM unavailable

**Anti-theatre rationale:** Dung extensions (grounded, preferred, stable, CF2, etc.) require Tweety AF framework. Python-side combinatorial enumeration has no soundness guarantee — honest-absent is strictly better than fabricated results written to analysis state.

## Test plan

- [x] FP-22 tests: 4/4 pass (`TestDungNativeSurvives`, `TestPythonDungFallback` x2, `TestDungInvokeFallbackSentinel`)
- [x] `TestABAFailLoud` failure confirmed pre-existing (fails on stash = unmodified code)
- [x] mypy strict on `invoke_callables.py`: 0 issues
- [x] Rebase on current main (up to date)

## Files removed and why

Per Cleanup-Gate policy (docs/proposals/CLAUDE_MD_CLEANUP_GATE_ADDENDUM.md):

| File | Type | Last useful date | Reason for deletion |
|------|------|-----------------|---------------------|
| _python_dung_fallback body (147 lines, kept as stub) | implementation | 2025 (original) | Theatre: produced fabricated extension sets without JVM (#1019) |

The function itself is preserved as a fail-loud stub — not deleted entirely — to prevent accidental invocations from future code paths.

Closes #1249. Epic #1191.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/argumentation_analysis/agents/core/logic/pl_handler.py
+++ b/argumentation_analysis/agents/core/logic/pl_handler.py
@@ -509,3 +509,181 @@ class PLHandler:
         )
         handler = self._get_sat_handler()
         return handler.query(normalized_kb, normalized_query, settings.pysat_solver)
+
+    # ── Multi-backend comparison (FP-20 #1244, mandate R468) ──────────
+
+    # PySAT backends that DECIDE firsthand (probe 2026-06-23, synthetic atoms):
+    # each returns the correct verdict on BOTH {P}->SAT and {P,!P}->UNSAT.
+    # ``cryptominisat5`` is EXCLUDED: it returns UNSAT on a trivially-SAT formula
+    # (``AttributeError: 'CryptoMinisat' object has no attribute 'cryptosat'`` —
+    # its native binding is broken in this env). Promoting a backend that emits a
+    # wrong verdict would fabricate a comparison point — worse than absent (#1019).
+    # A future env where cryptominisat5 genuinely decides can re-add it; the
+    # per-backend sentinel test guards the decision both ways.
+    PL_COMPARISON_PYSAT_BACKENDS: List[str] = [
+        "cadical195",
+        "glucose42",
+        "maplechrono",
+        "lingeling",
+        "minisat22",
+    ]
+
+    async def compare_pl_backends(self, knowledge_base_str: str) -> dict:
+        """Run ALL available PL/SAT backends on the same KB and compare verdicts.
+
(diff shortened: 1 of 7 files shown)


jsboigeEpita (commented):
**[Hermes]** COMMENT_WITH_CONCERNS

**Security**: clean.

**Agree with the fail-loud principle** (#1019 mandate): a silent Python fallback that computes Dung extensions when the JVM is unavailable masks infrastructure failures. Replacing `_python_dung_fallback` with a `RuntimeError` stub that surfaces "JVM/Tweety required" is the correct architectural choice — degraded-but-correct is worse than fail-loud when the canonical path is the JVM.

**Tests correctly updated**: `TestDungNativeSurvives` flipped to expect `RuntimeError` (was: expected real results). The new `TestDungInvokeFallbackSentinel` verifies `_invoke_dung_extensions` returns `degraded=True, extensions={}` when `af_handler` import fails — good sentinel coverage.

**Concern (semantic, non-blocking)**: the title says "kill theatre" but the removed `_python_dung_fallback` was mathematically exact for small AFs (grounded/complete/preferred/stable via brute-force enumeration is correct for small argument counts). It was not fabrication in the strict sense — it was a correct-but-degraded solver. The architectural tradeoff (fail-loud > silent fallback even when fallback is correct) is sound per #1019, but the framing could overstate what was removed. Minor — the decision itself is right.

No blockers. Clean removal, tests consistent.

PR #1322, opened 2026-07-01
fix(testing): move environment import below shebang in 6 scripts

jsboigeEpita:
Found via runtime-import scan (R517 dig). The auto_env migration `fd5bf726` bulk-prepended `import argumentation_analysis.core.environment` to L1 of entry-point scripts, pushing the shebang to L2/L3 — corrupting the shebang convention. Same defect class as the one that helped kill `run_embed_tests.py` (#1317/#1319).

## Root cause
`fd5bf726` "Migrate auto_env" prepended the env import at L1 instead of inserting it after the standard header (`shebang` → `coding cookie` → `docstring`). The goal (auto-activate conda early) is preserved by placing the import immediately after the header — the module self-executes on import either way, and the shebang/coding-cookie are just comments to the interpreter.

## Scope (6 scripts, pure reorder, +6/-12)
| File | Before | After |
|---|---|---|
| `debug_test_crypto_cycle.py` | env import L1, comment L3 | comment L1, env import L2 |
| `get_test_metrics.py` | env L1, shebang L3 | shebang L1, env L3 |
| `investigation_playwright_async.py` | env L1, shebang L3 | shebang L1, env L4 |
| `run_tests_alternative.py` | env L1, shebang L3 | shebang L1, env L3 |
| `simple_test_runner.py` | env L1, shebang L3 | shebang L1, env L3 |
| `validate_embed_tests.py` | env L1, shebang L3 | shebang L1, env L3 |

## Verification (firsthand, leçons R512/R515)
- **`py_compile` on all 6** → all compile (not assumed, verified).
- Environment auto-activation still loads early: the import is now the **first import** in each file, before any project import — the module's documented behavior ("s'auto-exécute à l'import") is preserved.
- `run_pytest.py` (same folder) already has the correct `#!/usr/bin/env python3` header at L1 → confirms this is the repo convention; the 6 affected scripts were the outliers from the bulk prepend.
- **Excluded `run_embed_tests.py`** — handled by #1319 (deletion, not reorder). The two PRs are disjoint (different files).
- mypy strict CI gate (8 core files) **unaffected** — all scripts are outside that scope; no production code touched, only header comments moved.

## Why this is a real defect (not cosmetic)
[...]

PR #1325, opened 2026-07-01
fix(scripts): remove duplicate environment imports in 2 scripts

jsboigeEpita:
## Summary

Continuation of the auto-env migration cleanup (`fd5bf726`). The shebang reorder sweep (#1322 testing/, #1323 bulk) noted one bonus finding — a duplicate env-import in `correct_source_paths.py` — left unbundled to keep those PRs focused. This PR closes that finding **and** extends the scan repo-wide: a second duplicate surfaced in `orchestrate_webapp_detached.py`.

`import argumentation_analysis.core.environment` auto-executes the module on first import (activates conda / sets up the path). A second identical import is a no-op — pure dead code. Same artifact family as #1322/#1323.

## Changes

| File | Before | After |
|------|--------|-------|
| `scripts/maintenance/tools/correct_source_paths.py` | identical import on L1 **and** L2 | keep one (L1) |
| `scripts/orchestration/orchestrate_webapp_detached.py` | L2 (correct, first statement after shebang) **+** L21 stale `# Import déplacé ici` (original never removed) | keep early L2, drop L21 |

## Verification (firsthand)

- **py_compile both OK** — no syntax regression.
- **1 env-import per file** (was 2) — auto-env still runs once, as the first statement.
- **2 deletions, 0 logic changed** — pure dead-code removal.
- For `orchestrate_webapp_detached.py`: kept the **early** import (L2, right after shebang = the convention #1322/#1323 restored), dropped the redundant "moved here" copy that ran after `sys.path` config.
- Scan was repo-wide (`grep -c` over `scripts/`): exactly these 2 files have count > 1; no others.

## Scope

Scripts are outside the 8-file mypy-strict CI gate, so no type-check impact. Docs/scripts-only, disjoint from all open PRs.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/scripts/maintenance/tools/correct_source_paths.py
+++ b/scripts/maintenance/tools/correct_source_paths.py
@@ -1,5 +1,4 @@
 import argumentation_analysis.core.environment
-import argumentation_analysis.core.environment
 import json
 import logging
 from pathlib import Path
--- a/scripts/orchestration/orchestrate_webapp_detached.py
+++ b/scripts/orchestration/orchestrate_webapp_detached.py
@@ -18,7 +18,6 @@ project_root = Path(__file__).resolve().parent.parent.parent
 sys.path.insert(0, str(project_root))
 
 # --- Imports du projet (après configuration du path) ---
-import argumentation_analysis.core.environment  # Import déplacé ici
 from project_core.service_manager import InfrastructureServiceManager, ServiceConfig
 
 


jsboigeEpita (commented):
Cross-verify (po-2023, read-only) — clean, no gap found.

Firsthand at PR head: both files have exactly 1 env-import (was 2), so auto-env still runs once. orchestrate_webapp_detached.py keeps the early L2 import right after shebang; only the redundant L21 stale copy is dropped. Verified via file-at-HEAD, not just the hunk. 2 deletions, 0 logic change. Disjoint from #1324. Safe to merge.

(Could not use --approve: the shared author account authored this PR; this comment stands as the cross-verify record.)

PR #1330, opened 2026-07-01
fix(examples): repair nested-f-string SyntaxError in spectacular demo

jsboigeEpita:
## What

`examples/02_core_system_demos/scripts_demonstration/demonstration_epita_spectacular.py` did **not compile** on Python 3.10 (the CI target) — `SyntaxError: f-string: unterminated string` / `unmatched '['`. The whole spectacular demo was unrunnable on the CI Python.

## Root cause (Python 3.10 vs 3.12, PEP 701 — same family as #1329)

11 sites reused the **outer f-string's quote char inside a nested f-string**, e.g.:
```
print(f"    {_c(conf_color, f'{claim["confidence"]:.0%}')}  ...")
```
The outer f-string is `f"..."` (double-quoted). The nested `f'{claim["confidence"]:.0%}'` contains `"` — which pre-PEP 701 (Python < 3.12) **terminates the outer string**, since f-string quote scanning was naive. Both double-quote (`claim["confidence"]`) and single-quote (`m['system']` inside `f'...'`) variants appear. Python 3.12+ (PEP 701) allows this, so it passed on the dev env (3.13) but fails on CI (3.10).

## Why fix, not delete

This is the **spectacular demo** (#361/#362/#377) — the Capstone-runner demonstration script, live and load-bearing. It also sits in the **same demos directory** as `demonstration_epita.py`, which PR #1328 (po-2023) points users to as the *recommended entry-point*. A broken sibling script made that entry-point dir misleading.

## Fix (anti-pendule: extract, don't requote)

Extract each nested dict access to a local variable before the f-string, eliminating the quote nesting entirely:
```
conf = claim["confidence"]
claim_text = claim["text"]
...
print(f"    {_c(conf_color, f'{conf:.0%}')}  {claim_text[:65]}")
```
No logic change, no behavioral change. Readable and idiomatic.

## Validation

- `python -m py_compile` -> **OK** (was SyntaxError).
- **End-to-end run** (self-contained mock demo, no API/corpus needed): `DEMONSTRATION COMPLETE`, 6 steps, 32 fields rendered, ANSI colors + all formatted values (confidences as %, scores `.1f`) intact -> runtime semantics verified, not just syntax.
- Repo-wide `py_compile` sweep (1146 files across tests/examples/project_core/interface_web/api/config/Arg_Semantic_Index/abs_arg_dung/CaseAI) -> **0 broken** (was 1).
- Disjoint from open PRs (#1327 modal, #1328 tutorials, #1329 interactive_guide — same 3.10/3.12 family but different files).

## Provenance

Found during R523 active-dig recon (no-idle mandate): extended the R522 `py_compile` sweep (which had covered only `scripts/` + `argumentation_analysis/`) to the remaining canonical dirs. R522 surfaced #1329; this is the `examples/` hit.

Co-Authored-By: Claude-Code <noreply@anthropic.com>

### DIFF
--- a/examples/02_core_system_demos/scripts_demonstration/demonstration_epita_spectacular.py
+++ b/examples/02_core_system_demos/scripts_demonstration/demonstration_epita_spectacular.py
@@ -310,15 +310,23 @@ def render_step_1(data: Dict):
     print()
     kv("Claims extracted:", str(len(out["extracted_claims"])), C.G)
     for claim in out["extracted_claims"]:
-        conf_color = C.G if claim["confidence"] >= 0.85 else C.Y if claim["confidence"] >= 0.70 else C.R
-        print(f"    {_c(conf_color, f'{claim["confidence"]:.0%}')}  {claim["text"][:65]}")
-        print(f"         {_c(C.DIM, f'type={claim["type"]}  id={claim["id"]}')}")
+        # Local vars avoid nested same-quote f-strings (PEP 701-incompatible on
+        # Python 3.10 = CI target): dict access with " inside f'...' inside f"..."
+        # terminates the outer string. See #1329 (same family).
+        conf = claim["confidence"]
+        conf_color = C.G if conf >= 0.85 else C.Y if conf >= 0.70 else C.R
+        claim_text = claim["text"]
+        claim_type = claim["type"]
+        claim_id = claim["id"]
+        print(f"    {_c(conf_color, f'{conf:.0%}')}  {claim_text[:65]}")
+        print(f"         {_c(C.DIM, f'type={claim_type}  id={claim_id}')}")
 
     print()
     kv("Arguments:", str(len(out["extracted_arguments"])), C.B)
     for arg in out["extracted_arguments"]:
         premises = ", ".join(arg["premises"])
-        print(f"    {premises} → {arg['conclusion']}  {_c(C.DIM, f'({arg["type"]})')}")
+        arg_type = arg["type"]
+        print(f"    {premises} → {arg['conclusion']}  {_c(C.DIM, f'({arg_type})')}")
 
     print()
     kv("Extraction time:", f"{out['extraction_stats']['time_ms']}ms", C.DIM)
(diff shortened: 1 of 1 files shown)


jsboige (commented):
Cross-verify (po-2023, read-only) — VALIDATED, safe to merge.

Firsthand checks:
- **py_compile + AST parse at PR head = OK** (no typo introduced by the refactor).
- **Semantic trace of all 11 extraction sites = identical behavior.** Each pattern is `local = d["k"]; ...f'...{local}...'` replacing `f'...{d["k"]}...'` — the local holds the same value, same format spec, same slicing. Traced each hunk: conf/claim_text/type/id (step1), arg_type, natural (step2 FOL), m_system/m_meaning (step2 modal), sev (step3), atk_type (step4 attacks), bstatus/retracted_by (step4 JTMS), round_num/score_def/score_atk/ca_strength (step5). No logic change anywhere.
- **The retracted_by extraction is correctly INSIDE the `if "retracted_by" in bdata:` guard** (assigning it outside would KeyError when the key is absent). Verified placement.
- Sites LEFT inline (e.g. `arg['conclusion']`, `m['formula']`, `r['defender']`) are single-quote access inside a double-quoted outer f-string = NO PEP701 conflict = correctly left (not over-refactored, anti-pendule respected).

Honest caveat: I tried an AST nesting-depth checker to prove "0 residual same-quote-nested f-strings", but it has a blind spot — it only descends FormattedValue→FormattedValue chains and misses JoinedStrs nested inside Call args (which is exactly this file's `f"...{_c(c, f'{x["k"]}')}"` pattern). So the "0 nested" result is a false negative from my checker, NOT independent proof of completeness. I relied on the manual diff trace instead, which confirms every genuine same-quote conflict was extracted and every left-inline site is genuinely safe.

Same PEP701 family as #1329 (CI = Python 3.10, dev = 3.13 where the bug is invisible). Load-bearing script (capstone runner + sibling of the #1328-recommended entry-point). Disjoint from #1327/#1328/#1329. Safe to merge.

PR #1681, opened 2026-08-07
fix(aspic): writer must carry the qualified attacks list (#1649 post-#1679)

jsboigeEpita:
## #1649 — ASPIC+ writer must surface the qualified attacks list

Resolves the writer-side loss exposed by #1679 (the ASPIC+ feeding PR). The handler now produces a qualified attacks list with structural scope (`undercut` / `rebut` / `undermine` / `unresolved`), but the writer dropped it on the floor: `state.aspic_results[0]` had no `attacks` field, the axis projected as a Dung copy, and the unique singular contribution of ASPIC+ (#1649) was invisible to any reader of the state.

### The fix — two coordinated edits

**1. `core/shared_state.py:add_aspic_result`** — extend the signature with `attacks: Optional[List[Dict[str, Any]]] = None` (default keeps all 3-arg callers backward compatible: SK plugin wrappers, tests, fixtures). The new field is serialized as `[]` when `None` so downstream readers never see the distinction between "handler ran, no attacks" and "writer dropped the field".

**2. `orchestration/state_writers.py:_write_aspic_to_state`** — read `output.get("attacks")`, normalize non-list shapes to `None` (defensive — a malformed value should not crash the writer), pass to `add_aspic_result(reasoner_type, extensions, statistics, attacks=attacks)`.

### Anti-pendule

- **Top-level entry field, not a `formalism_specific` sidecar.** The ABA writer (site 1 of #1648 Wave-2, PR #1680) uses the sidecar pattern because `dung_frameworks` has a constrained `arguments/attacks/extensions` projection. `aspic_results` is a flat list with explicit peer fields (`extensions`, `statistics`) — `attacks` is a peer, the top-level shape is the natural one. Recording the design choice in the docstring so the next site-by-site fix picks the right shape.
- **SK plugin signatures untouched.** `state_manager_plugin.py` and `phase_scoped_state.py` keep their 3-arg `add_aspic_result` wrappers — the LLM does not write attacks, the writer does, post-handler. The LLM-facing surface is preserved.
- **11 readers of `aspic_results` not migrated** — sanitize_state, restitution, deep_synthesis, appendix, multi_format_exporter, reprompt_trace, tweety_result_interpretation_plugin, conversational_orchestrator, conversational_benchmark, … A reader that wants the field reads `entry["attacks"]`. Anti-pendule: additif, single site, no reader migration.

### Privacy

Synthetic atoms only in tests (`r_main`, `r_neg_main`, `concl_x`, `p`, `q`, `a`, `b`). **Known gap** (out of scope, follow-up): `sanitize_state.py` does not yet opacify `aspic_results[*].attacks.attacker_premises` or `.attacker_rule` — these may carry corpus tokens in production (LLM-named rules that survive id validation in #1679's translator). The sanitiser update is its own task; the coord's anti-pendule rule forbids addressing here.

### Anti-#1019 discipline (R761/R764/R765/R767)

- **Real writer, handler stub.** `test_state_writers_1649.py` drives `_write_aspic_to_state` directly (no mocked writer) and stubs the handler output to mirror the post-#1679 shape. A mocked writer would just agree with itself.
- **Differential proven.** With the writer edits reverted (5 tests FAIL — `writer dropped the qualified attacks list`, `writer dropped attacks shape`, etc.); with the edits in place (5 tests PASS — `attacks` flows to state with all 3 scopes preserved).

### DoD checklist (the writer-side line items of #1649)

- [x] The entry `state.aspic_results[-1]` carries a `attacks` list paralleling `extensions` and `statistics`.
- [x] Each attack preserves `{attacker_rule, attacker_premises, target, scope}`.
- [x] All 4 ASPIC+ scopes (`undercut` / `rebut` / `undermine` / `unresolved`) survive the writer boundary (the 3 non-`unresolved` scopes pinned by separate assertions).
- [x] Empty handler `attacks` serializes to `[]` (not `None`, not missing).
- [x] Malformed `attacks` (non-list) normalizes to `None` → `[]`.
- [x] `add_aspic_result(reasoner_type, extensions, statistics)` — the legacy 3-arg form — still works and now defaults `attacks` to `[]`.
- [x] **No reader migration** — sanitize_state, restitution, appendix, etc. unchanged.
- [x] **SK plugin signatures unchanged** — LLM-facing surface preserved.

### Validation

- **mypy strict** on the 2 modified source files: **0 issues**.
- **`test_state_writers_1649.py`** (new): **5 passed**.
- **`test_state_writers_1648.py` + `test_unified_pipeline.py`** (regression): **169 passed, 9 xfailed** (unrelated sites).
- **Differential** (stash writer → run new tests): 5 FAIL; pop stash → 5 PASS.

### Scope — writer-side only

This PR closes the writer-side line of #1649. The reader-side (prose channel for ASPIC+ arbitration, `#1667` presence channel at 2/5 axes) is a separate task — a reader over an inert axis is theatre #1019, and the writer-side prerequisite is exactly what this PR delivers.

🤖 Worker myia-po-2025 — R769 (ASPIC+ writer-side, post-#1679)

### DIFF
--- a/argumentation_analysis/core/shared_state.py
+++ b/argumentation_analysis/core/shared_state.py
@@ -954,15 +954,35 @@ class UnifiedAnalysisState(RhetoricalAnalysisState):
         return rk_id
 
     def add_aspic_result(
-        self, reasoner_type: str, extensions: List[Any], statistics: Dict[str, Any]
+        self,
+        reasoner_type: str,
+        extensions: List[Any],
+        statistics: Dict[str, Any],
+        attacks: Optional[List[Dict[str, Any]]] = None,
     ) -> str:
-        """Add an ASPIC+ analysis result."""
+        """Add an ASPIC+ analysis result.
+
+        #1649 (#1678 #1679): ``attacks`` carries the qualified attack list
+        produced by ``ASPICHandler.analyze_aspic_framework`` — a list of dicts
+        shaped ``{attacker_rule, attacker_premises, target, scope}`` with
+        ``scope`` in ``{undercut, rebut, undermine, unresolved}``. The field
+        is the singular contribution of ASPIC+ (#1649): without it the
+        projection is a Dung copy and the axis loses its only reason to
+        exist. ``attacks`` defaults to ``None`` so all 3-arg callers
+        (SK plugin wrappers, tests, fixtures) stay backward compatible;
+        ``None`` is serialized as the empty list so readers never see the
+        distinction. The Semantic Kernel plugin wrappers
+        (``state_manager_plugin.py``, ``phase_scoped_state.py``) keep the
+        3-arg signature by design — the LLM does not write attacks, the
+        writer does, post-handler.
+        """
         as_id = self._generate_id("aspic", self.aspic_results)
         entry = {
             "id": as_id,
             "reasoner_type": reasoner_type,
             "extensions": extensions,
             "statistics": statistics,
+            "attacks": list(attacks) if attacks is not None else [],
         }
         self.aspic_results.append(entry)
         state_logger.info(f"ASPIC+ result added: {as_id} (reasoner: {reasoner_type})")
--- a/argumentation_analysis/orchestration/state_writers.py
+++ b/argumentation_analysis/orchestration/state_writers.py
@@ -821,18 +821,39 @@ def _write_ranking_to_state(output: Any, state: Any, ctx: dict[str, Any]) -> Non
 
 
 def _write_aspic_to_state(output: Any, state: Any, ctx: dict[str, Any]) -> None:
-    """Write ASPIC+ analysis results to UnifiedAnalysisState."""
+    """Write ASPIC+ analysis results to UnifiedAnalysisState.
+
+    #1649 (#1678 #1679): the handler now produces a qualified ``attacks``
+    list (undercut / rebut / undermine / unresolved, scoped structurally on
+    the rule inputs). Pre-fix the writer dropped it on the floor — ASPIC+
+    rendered as a Dung copy and the axis lost its singular contribution.
+    We surface ``attacks`` as a top-level entry field (peer of ``extensions``
+    and ``statistics``), NOT via a ``formalism_specific`` sidecar: the
+    ABA writer (site 1 of #1648 Wave-2) uses the sidecar pattern because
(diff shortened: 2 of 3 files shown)


jsboigeEpita (commented):
## Integration contract verified (as #1678 author)

Read the diff + body before commenting. The writer correctly consumes the handler's #1679 output shape — I built `_qualify_attacks`, so I can confirm point by point:

- `output.get("attacks")` reads the exact key the handler emits ✓
- The test stub mirrors the real keys — `{attacker_rule, attacker_premises, target, scope}` with `scope ∈ {undercut, rebut, undermine, unresolved}` ✓ (these are my exact field names)
- `None` → `[]` serialization, non-list → `None` normalization, 3-arg backward compat all look right ✓

The writer-side fix is correct and the DoD items check out.

## One refinement on the flagged privacy gap (severity downgrade, not a blocker)

The "Known gap" says `attacker_premises` / `attacker_rule` "may carry corpus tokens in production (LLM-named rules that survive id validation)". I think that's **more severe than the production path warrants** — worth recalibrating so the follow-up isn't over-prioritized:

- **`attacker_rule`** = the rule's `name`, which in production goes through `_sanitize_rule_name` → `def_{lowercased_alnum}[:48]` (lossy, opaque). Not raw text.
- **`target` / `attacker_premises`** = the rule's head/body atoms. In production the translator runs every head/body through `_pl_atom` (`invoke_callables.py:3734`), which is lossy opaque: ≤24 leading alnum chars, lowercased, non-alnum collapsed to `_`, plus a short hash suffix. Its own docstring cites the `teleprompteranecdote` leak case as exactly what it exists to prevent.

So the residual risk is a **≤24-char lowercased alnum prefix fingerprint** (distinctive substrings like `unknownobjecttelepromp` could in theory identify a source), **not verbatim corpus tokens**. Closing the `sanitize_state` gap is still worth doing as defense-in-depth, but it's lower urgency than "corpus tokens leak into state/dashboard".

**Caveat worth recording for the follow-up:** the handler is a public API (`analyze_aspic_framework(strict_rules, defeasible_rules, axioms)`) and does **not** re-sanitize — the opaqueness is a property of the production caller (the translator), not a guarantee of the handler output. A direct caller passing raw atoms would leak verbatim. So the sanitizer is the right place to close it (not the handler).

No change requested — flagging so the coord sizes the follow-up correctly. 🤖

